Issue #119, opened 2018-01-12
Support Bitcoin sign


KrasnayaPloshchad:
It would be awesome if you add support for Bitcoin (₿) sign.

waldyrious:
Reference links:
- https://unicode-table.com/en/20BF/
- http://www.fileformat.info/info/unicode/char/20BF/

also, you misspelled "Bitcoin" as "Bitcion" in the issue title :)

KrasnayaPloshchad:
OK, I fixed it.

PR #125, opened 2018-02-18
Issue #119: Add Bitcoin currency glyph to Libertinus serif

mgavioli:
This adds the Bitcoin glyph to the serif fonts; sans fonts will follow, if the PR is OK.

This is my first contribution to this project, I hope I got it right; I also do not know if the .OTF files have to be re-generated with each PR or if they are generated only upon release.


khaledhosny:
Looks good, thanks! Generating the fonts is preferred (using `make` not from FontForge GUI), but I can do that if it is too much trouble.

For the glyph, I prefer to keep the overlapping outlines in the `Fore` layer and mark the glyph for overlap removal at build time (Element → Glyph Info… → Mark for Unlink, Remove Overlap before Generating), this would also make the outlines in `Back` layer moot.

You should also use `tools/sfdclean.sh` to clean some of the cruft FontForge adds to SFD files (my bad, I never committed or documented this before).

khaledhosny:
I took the liberty of doing these changes and push manually.

mgavioli:
Thanks for 'completing' the PR. I'll try to follow your indications for the next ones (over time, I'll try to get rid of all the issues I can understand / deal with).

I work with Linux Mint 17.3 (derived from Ubuntu Trusty); am I correct that, in order to use the supplied  `makefile` I need to install:
- python-fontforge
- fontTools

Thanks!

khaledhosny:
> work with Linux Mint 17.3 (derived from Ubuntu Trusty); am I correct that, in order to use the supplied makefile I need to install:
>
> *    python-fontforge
> *    fontTools

I think so, though I use the master version of FontForge so the final output might vary slightly, but that shouldn’t be a real issue. I also suggest installing fontTools using pip, your distro version might be few years old.

mgavioli:
> I use the master version of FontForge so the final output might vary slightly, but that shouldn’t be a real issue.

I use the latest release from Launchpad PPA (currently July 2017), which is not the latest dev, but for production tools I prefer to use releases considered stable by their dev team. As you say, output should not be so different, though.

khaledhosny:
An easy way to see if your setup is producing OTF identical to mine is to run `make -B` on an unmodified master, if there is no change in the OTF files then we are good, otherwise I’ll just generate them myself (if there are any changes, then diff the output of `ttx` for the old and new files should give some clues).

mgavioli:
@khaledhosny : Sorry, I didn't notice this post! I run the test you suggested:

1. backed the current master .OTF's up
2. run `make -B` on current mater
3. generate `ttx` output from each backed-up file and each locally generated .OTF's
4. diff each pair

There are differences, but they are only in the `<head>` section and the same 3 for all fonts:

- Original:  &lt;flags value="000000**0**0 00001011"/&gt;
   Local:    &lt;flags value="000000**1**0 00001011"/&gt;

- Original: &lt;fontDirectionHint value="**2**"/&gt;
   Local:    &lt;fontDirectionHint value="**0**"/&gt;

- and of course &lt;checkSumAdjustment value="[variable]"/&gt; which depends on the others.

Does this qualify as an important enough difference to prevent using my July 2017 version of FontForge for this project?

khaledhosny:
I already changed the build script to normalize this, see d2c73f93221f037672396afbe0f6935323287f32, so your build should be identical to mine now (and no, that wasn’t an important change, but it is nice to have reproducible build, it should help catching tool bug as well).

PR #138, opened 2018-03-09
Fixes #98 : J and J.alt

### LINKED ISSUE #98: In Libertinus Serif, J looks different in bold italic face.
Also see: https://sourceforge.net/p/linuxlibertine/bugs/326/

mgavioli:
Makes uniform `J` 'swashy' shape and `J.alt` straight shape in Serif italics and make `J.alt` accessible via `salt` feature.
The 'swashy' `J.alt` glyph in Serif Regular has been left and make accessible via `salt`.

Fixes https://github.com/khaledhosny/libertinus/issues/98

Summary:

- **Sans Regular**: removed incongruous `J.alt` glyph (this caused all following glyphs to change ID and a number of references to change);
- **Sans Italic**: removed incongruous `J.alt` glyph (this caused all following glyphs to change ID and a number of references to change);

- **Serif Regular**: added `J.alt` (swashy) to `salt` lookup as alternate of `J` (straight);
- **Serif Italic**: added `J.alt` (straight) to `salt` lookup as alternate of `J` (swashy);
- **Serif Semi-bold it.**: drawn new `J.alt` (straight) as 75% interpolation between Italic and Bold italic `J.alt`,
added `salt` lookup table,
added `J.alt` (straight) to `salt` lookup as alternate of `J` (swashy);
- **Serif Bold italic**: re-drawn `J` as bold of Italic `J` (swashy),
added `J.alt` (straight) to `salt` lookup as alternate of `J` (swashy).

LibreOffice screen shot:

issue_098_screenshot_02


waldyrious:
Was it intentional to keep `J.alt` for serif regular and not add one for semi-bold and bold? I think both options that would make this more consistent (having `J.alt` for all three, and not having `J.alt` for any of the three) would remain aligned with the comments by @khaledhosny in #98:

> I'm OK with serif roman not having an alternate form (but it is OK to have one).

> In general I don’t think we should change the defaults too much

(Although, as a side note, I agree with you that it would be even more consistent to have the "swashy" shape always be the alt glyph. I don't think it would be deviating too much from the original font, but rather resolving a styling inconsistency already present in it)

khaledhosny:
Please fix the `make check` failure and I’ll merge (you need to drop features that reference removed glyphs).

mgavioli:
Opps! I expected FF to do it automagically upon deleting a glyph. But in some sense it does it: the .OTF generated after removing the spurious reference had no difference with the previous version. Pushing the change...

mgavioli:
@waldyrious : yes, it was (more or less) intentional: I didn't want to loose the existing Regular `J.alt` shape (as with the pork: one does not throw anything away!), but I was not sure (semi-)bold versions of it would really make sense. So, eventually, yes, there are two empty slots in the grid above; symmetry of life is not always self-evident...

waldyrious:
I can understand your reasoning, @mgavioli; thanks for clarifying.

PR #146, opened 2018-03-29
Fix #115 : Celsius / Fahrenheit glyphs

### LINKED ISSUE #115: Degree Celsius/Fahrenheit signs not bolded in Libertinus Serif Bold Italic
Degree Celsius and Degree Fahrenheit signs looks the same in Libertinus Serif Italic and Libertinus Serif Bold Italic.

See: #112

mgavioli:
Fix #115 : Celsius / Fahrenheit glyphs

Fixes https://github.com/khaledhosny/libertinus/issues/115

Summary:

The 'C' and 'F' parts of these glyphs have been replaced with references to the main 'C' or 'F' glyphs scaled to 89%.
The ring shape have been replaced with a reference to 'degree' (when not already so).

Detail:

**Regular**:
- Celsius: replaced reference to U+21&d (Roman numeral 100) with scaled ref. to main 'C'
- Fahrenheit: replaced 'F' shape with scaled ref. to main 'F'

**Italic**:
- Celsius: replaced 'C' shape with scaled ref. to main 'C' and ring shape with reference to `degree`
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' and ring shape with reference to `degree`

**Semi-bold**:
- Celsius: replaced 'C' shape with scaled ref. to main 'C' (ring shape was already a reference to `degree`)
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' (ring shape was already a reference to `degree`)

**Semi-bold italic**:
- Celsius: replaced 'C' shape with scaled ref. to main 'C' (ring shape was already a reference to `degree`)
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' (ring shape was already a reference to `degree`)

**Bold**:
- Celsius: replaced 'C' shape with scaled ref. to main 'C' (ring shape was already a reference to `degree`)
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' (ring shape was already a reference to `degree`)

**Bold italic**:
- degree: made really bold italic
- Celsius: replaced 'C' shape with scaled ref. to main 'C' and ring shape with reference to `degree`
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' and ring shape with reference to `degree`

**Display**:
- Celsius: replaced 'C' shape with scaled ref. to main 'C' and ring shape with reference to `degree`
- Fahrenheit:  replaced 'F' shape with scaled ref. to main 'F' and ring shape with reference to `degree`


mgavioli:
Fixing for Travis check...

khaledhosny:
Can you use scaled references of the C and F glyphs? This should keep the shapes in sync.

mgavioli:
In principle, yes, scaled references could be used. In practice, I have learned to avoid them, as some rasterisers choke on them. Or at least, they used to; maybe this limitation has been overcome in the meantime (my experience on this point is rather ancient), but I would consider it risky.

It would be possible to mark the glyphs for unlink; but this generates a number of non-integral point coordinates which make the font bigger and rasterisation slower (in my shapes described above, all coordinates have been rounded to integer).

Anyway, if you think it would be preferable nevertheless, I'll modify the fonts.

khaledhosny:
> In principle, yes, scaled references could be used. In practice, I have learned to avoid them, as some rasterisers choke on them. Or at least, they used to; maybe this limitation has been overcome in the meantime (my experience on this point is rather ancient), but I would consider it risky.

This might be true for TrueType outlines, but CFF does not have references (AKA composite glyphs), thus is purely an editor thing, the generated font will be equivalent either way.

> It would be possible to mark the glyphs for unlink; but this generates a number of non-integral point coordinates which make the font bigger and rasterisation slower (in my shapes described above, all coordinates have been rounded to integer).

I wouldn’t worry about this, and if we want to always round, we should to it in `tools/build.py` not in the sources.

mgavioli:
OK then. I'll redo the modifications.

mgavioli:
Replaced explicit outlines with scaled refs. to 'C' and 'F'.

khaledhosny:
Thanks!

Issue #40, opened 2016-06-13
Libertinus uses alternative glyphs for any numerals whose size are less than the base size


taikis:
I have also posted about this issue on [tex.SE](http://tex.stackexchange.com/q/314535/29127).

For some reason, in the math mode, alternative glyphs are used for any numerals whose size is less than `\normalsize`, explicitly (that is, through `\small`, `\scriptsize`, `\footnotesize`, and `\tiny)` or implicitly (for example, through `^` and `_`), leading to unpleasant kerning for consecutive numerals etc etc. Is this an intended behaviour?

Currently I wrap all sub/superscripts in `\mathrm` as a workaround.

```
\documentclass{article}
\usepackage{unicode-math}
\setmainfont{Libertinus Serif}
\setmathfont{Libertinus Math}

\begin{document}

$1234567890^{1234567890}$

$1234567890^\mathrm{1234567890}$

\end{document}
```

[![1234567890^1234567890](http://i.stack.imgur.com/Uto2O.png)](http://i.stack.imgur.com/Uto2O.png)

khaledhosny:
They are supposed to be used only for `\scriptsize` and `\scriptscriptsize`, if they are used anywhere else then it is an issue with LaTeX or the respective packages.

But I welcome any comment on the number themselves, they are based on Linux Libertine alternate glyphs for `sup` and `sub` OpenType features as I thought they are a good fit for the smaller sizes being designed for such usage, but may be they need more work to get them into shape.

taikis:
Ah OK. It's fine since it's an intentional design choice; I shall close this issue then. But personally I feel that uniform glyphs for all sizes provide better look, especially in a maths/physics document.

khaledhosny:
I might as will just kill them and use the regular digits, they look too intrusive to my liking actually (or may be just keep them for `\scriptscriptsize` where they might be less intrusive).

taikis:
I vote for the regular glyphs for all sizes. This would eliminate LaTeX issues in which the alternative glyphs are picked up for sizes other than `\scriptscriptsize` (or whatever is specified).

khaledhosny:
I think the problem might be is that Libertinus sets a bigger script sizes than traditional TeX (8pt and 6pt, respectively, compared to 7pt and 5pt). Here is a comparison between different combinations. I think if we set smaller script sizes they will look better, with no optical alternates at all they look much lighter:
[...]

PR #149, opened 2018-03-30
Fix #113 - Service Mark and Trade Mark in Sans

### LINKED ISSUE #113: Service Mark and Trade Mark Sign has wrong from in Libertinus Sans
Service Mark (U+2120) and Trade Mark Sign (U+2122) appear to be adapted from the serif face in Libertinus Sans.

See: #112

mgavioli:
Fixes https://github.com/khaledhosny/libertinus/issues/113

In the three Sans fonts, the Service Mark (U+2120) and Trade Mark glyphs (U+2122) were not in the same style of the font and not sans at all.

They have been re-done using references to the main Latin 'M', 'S' and 'T' glyphs scaled to 54%. Spacing has also been revised, generally increased.


khaledhosny:
For small marks like these, wouldn’t it be better if they were emboldened a bit, no not look washed up next to normal sized glyphs? Can you attach before and after screenshots (and may be the serif too, for comparison)?

mgavioli:
Some screen shots from LibreOffice. I made two variants, one simply scaled and one scaled and then emboldened (approx. 1/3 of stem and bar widths).

**Original situation** (before any change):
issue_113_screenshot_original

**Scaled references**:
issue_113_screenshot_scaled

**Scaled and emboldened**:
issue_113_screenshot_scaled_embold

**Serif** (for comparison):
issue_113_screenshot_serif

Notes:
- in the Serif bold, `TM` and `SM` are not emboldened.
- in the Sans (italic and bold), there seems to be a spacing issue in the `tsiv` sequence.

khaledhosny:
Scaled and emboldened version looks better IMO, the scaled-only version looks too light.

nsemrau:
Subtly bolder super- and subscript characters seem to be a hallmark of high-quality typography (see e.g. [1 (caption of first image)](https://en.wikipedia.org/wiki/Subscript_and_superscript), [2]( https://www.fonts.com/content/learning/fontology/level-3/numbers/superior-and-inferior-figures)), so I vote for the emboldend version, too.

BTW: Due to the argument that slightly bigger is (percieved as) higher quality, I would also love to see a return of the alternative figures in issue #40 in the long run. :)

mgavioli:
Bolder version pushed.

I hope not to have messed things up squashing the commits of the two versions...

khaledhosny:
Thanks!

PR #151, opened 2018-04-09
Fix #145 - Cyrillic small letter ef in Display font

### LINKED ISSUE #145: Any reference about cyrillic small eff in Libertinus Serif Display?
Left: Libertinus Serif. Right: Libertinus Serif Display
Never seen any font like this. I suggest changing it

mgavioli:
Fixes https://github.com/khaledhosny/libertinus/issues/145

In the Display font, the gliph U+0444 (CYRILLIC SMALL LETTER EF) had non-standard top and bottom serifs. This PR replaces them with shapes modelled on the Serif Regular U+0444 serif shapes.

Additionally, the original U+0444 glyph was moved to an un-encoded slot, named `uni0444.alt`, as the Serif Regular font also does. Neither in the Serif Regular nor in Display this alternate glyph is referenced by any lookup table.

See below for a screen shot.


mgavioli:
@khaledhosny : really OT here, but I cannot think of a better place.

With this PR, all listed bugs for Latin not having to do with diacritical positioning should be done (except https://github.com/khaledhosny/libertinus/issues/42 for which I am not very competent).

I now plan to address the diacritical positioning as a a whole, starting from the Serif Regular font. As this is going to take a while and it is likely to change a lot of glyphs, which may raise merge issues, I plan to PR each font separately, to minimise the delay, even if several issues will be fixed only by completing several fonts.

If you have comments on this, please advise.

khaledhosny:
> See below for a screen shot.

Did you forget to upload the screen shot?

> I now plan to address the diacritical positioning as a a whole, starting from the Serif Regular font. As this is going to take a while and it is likely to change a lot of glyphs, which may raise merge issues, I plan to PR each font separately, to minimise the delay, even if several issues will be fixed only by completing several fonts.

Doing it by font seems good.

mgavioli:
I am pretty sure I copied the link from the issue post. I just re-did it; let's see if this time it sticks.

khaledhosny:
Thanks!

KrasnayaPloshchad:
So which feature can be used to switch original U+0444 glyph?

FloraCanou:
I think there should be no feature to switch to it now, but one way to get it is to select through the special characters table in word processors like LO.

PR #176, opened 2018-05-08
Fix #162 : Names of A/O/Udieresis and A/O/Udieresis.alt

### LINKED ISSUE #162: Inconsistency in naming of some glyphs with diæresis?
**Situation**:
While revising the diacritics / anchor situation (currently for Serif Regular), I noticed that the 3 glyphs Ä (U+00C4), Ö (U+00D6), Ü (U+00DC) have unusual names: `Adieresis.alt`, `Odieresis.alt` and `Udieresis.alt` respectively.

They also bear a comment in German which, translated to English sounds like: ”These are alternative umlauts which are preferable in German".

The font also has glyphs named `Adieresis`, `Odieresis` and `Udieresis` (without `.alt`) in the unencoded slots:

issue_aoudieresis

To my eye, the `.alt` glyphs look totally normal, while the non-`.alt` version look 'stylized' and I remember similar shapes -- with more 'embedded' umlauts -- in German publications from the early-to-middle XX century.

The `.alt` forms (with Unicode encoding) are the ones shown by, for instance, LibreOffice without any special command or formatting (in any language I tried, including, but not limited to, German):

Finally, the `ss01` style set turns the `.alt` forms into the non-`.alt` forms (but not vice versa).

The Italic shows the same setup; I have not checked the other fonts, but two fonts are enough to tell this was intentional (at least in origin).

**Issue**:
This probably does not creates problems to the final user, but it breaks the automatic generation of accented glyphs (`Adieresis` references the `A.alt` glyph as base, rather than `A`) and will surely puzzle any future maintainer of the font(s).

**Suggestion**:
Move the `.alt` suffix from the names of the 'regular' glyphs (the top 3 in the image) to the names of the 'stylized' ones (the bottom 3), adapting the names in the `ss01` set.

I see no side effects in this; if accepted, I can include this in the forthcoming PR with updated anchor system.

mgavioli:
Fixes https://github.com/libertinus-fonts/libertinus/issues/162

Glyphs U+00C4, U+00D6 and U+00DC with the regular shapes for A/O/U+dieresis are named `Adieresis.alt`, `Odieresis.alt` and `Udieresis.alt`, while alternate glyphs with lower dieresis in unencoded slots are named `Adieresis`, `Odieresis` and `Udieresis`.

This PR swaps the names for a more consistent naming.

The Serif Bold-italic was different, as the names were correct but the shapes were exchanged; for this font, the PR swaps the shapes rather than the names.


sources/libertinussans-italic.sfd
```diff
@@ -189,7 +189,7 @@ OtfFeatName: 'ss02' 1033 "Swashy 'K', 'R'"
 OtfFeatName: 'ss03' 1033 "'double s' to two 's'"
 OtfFeatName: 'ss04' 1033 "'ampersand'"
 OtfFeatName: 'ss07' 1033 "Swap 'Eng' forms"
-Encoding: UnicodeBmp
+Encoding: Custom
```
khaledhosny: This should be kept as `UnicodeBmp`.

sources/libertinussans-bold.sfd
```diff
@@ -8099,7 +8099,7 @@ SplineSet
  165 647 138 672 138 703 c 0
 EndSplineSet
 Refer: 24 65 N 1 0 0 1 0 0 2
-Position2: "'cpsp' Versalabstaende 1" dx=2 dy=0 dh=5 dv=0
+Position2: "'cpsp' Versalabstaende 1" dx=2 dy=0 dh=5 dv=0 [ddx={} ddy={} ddh={} ddv={}]
```
khaledhosny: Those empty ` [ddx={} ddy={} ddh={} ddv={}]` additions are just adding noise, please remove them manually for now (I’ll see if I can update the normalizer to handle them).

sources/libertinusserif-bold.sfd
```diff
@@ -174,7 +174,7 @@ GaspTable: 3 8 2 17 1 65535 3 0
 OtfFeatName: 'ss01' 1033 "Low diaeresis on 'A', 'E', 'O'"
 OtfFeatName: 'ss03' 1033 "'double s' to two 's'"
 OtfFeatName: 'ss07' 1033 "Swap 'Eng' forms"
-Encoding: UnicodeFull
+Encoding: Custom
```
khaledhosny: Same here.

mgavioli:
Done. I have no idea where these changes came from, as they in some case involve glyphs I have not modified and, for sure, I have not modified the encoding.

khaledhosny:
> I have no idea where these changes came from, as they in some case involve glyphs I have not modified and, for sure

FontForge tends to touch glyphs even if one does very innocent things like previewing a substitution that involves this glyph!

> I have not modified the encoding.

Just renaming encoded glyphs can cause FontForge to switch to “Custom” encoding.

Issue #173, opened 2018-05-05
Documentation of Libertinus's open type features?


JakobWerner:
Maybe I just didn't search correctly, but I couldn't find a place where Libertinus's open type features, ligatures, stylistic sets, … are documented for the end user. Maybe a documentation (e.g. in the style of `texdoc junicode`) could be added? I would be willing to contribute one myself, but unfortunately I don't know about the available features myself, and also I don't think I could produce such a neat document.

khaledhosny:
We don’t currently have any user documentation, unfortunately. Some features are documented in the old Linux Libertine website: http://libertine-fonts.org/, and the definite list of features supported by each font can be obtained by `otfinfo`. If someone can contribute some user documentation, that’d be highly appreciated. I can answer any questions and help review it.

RGB-es:
I wrote a quick review for my blog (not sure if it's right to put a link to it here). As Khaled said, at least on Linux you can use 

`otfinfo -f libertinusserif-regular.otf`

to get a list of OT tags available and then just use a recent version of LibreOffice to test those features (or document them!). To test Libertinus Math you need to use XeTeX, though (I wrote about that, too...).

khaledhosny:
I moved the repository to its own organization for better organization (pun intented), and we now have a empty GitHub pages repository: https://github.com/libertinus-fonts/libertinus-fonts.github.io, in case someone wants to populate it with some useful documentation pages.

waldyrious:
> Some features are documented in the old Linux Libertine website: http://libertine-fonts.org

For reference, the latest archive.org snapshot of the old website at linuxlibertine.org is from [2017-10-14](http://web.archive.org/web/20171014200644/http://www.linuxlibertine.org:80/index.php?id=1&L=1).

KrasnayaPloshchad:
The following archive is good to be reference:
https://web.archive.org/web/*/http://www.linuxlibertine.org:80/fileadmin/user_upload/PDF/Libertine-XeTex-EN.pdf

gyuris:
Can you document '[nalt](https://docs.microsoft.com/en-us/typography/opentype/spec/features_ko#a-namenalt-idnaltatag-nalt)' feature, please? I see in the source (①-⑳) and with `otfinfo` but  I can't use it.

kenmcd:
The **nalt** feature does not seem to work how you would expect.  
You think it is going to take these three characters "(1)" and convert it to ① (U2460).
[...]

PR #182, opened 2018-05-20
Documentation for OpenType features, Libertinus Serif

gyuris:
Reaction on #173.


nsemrau:
There seems to be a miniscule typo in the `-ccmp` feature: "Shulfächer" should be "Schulfächer" to be orthographically correct German. :)

khaledhosny:
> 1. ss07: It doesn't works for me. 

It works on capital Eng Ŋ.

> 2. fina: σ→ς doesn't works for me.

This feature should be dropped; `fina` is to be only used for Arabic and similar scripts and most layout engines will not apply it to other scripts. In case of Greek, the two variants are encoded and should not be handled by layout features at all.

> 3. mkmk: I don't understand this feature.

This handles combining marks over other combining marks over other combining marks, e.g. x̆́.

> 4. aalt: Hm... Linux Libertine O works for me, but Libertinus doesn't.

I dropped that feature, it is rather useless IMO.

> 5. I can't reproduce CO₄³⁻ (see last rows in libertinus-sample.odt) with OpenType features...

I dropped the kerning here because it was non-standard, superscript and subscript modifier letters are phonetic symbols and not to be used in math and chemical formulas, these need specialized layout engines.

gyuris:
Thank you for the answer, Khaled. I will update my document shortly.

gyuris:
OK, @khaledhosny If you accept this documentation, please merge. I think there are no more known feature.

Please, say something about this feature documentation. 
If you want, I can add italic, sans, mono etc samples later.

gyuris:
OK, I'm done.

- File format: [.fodt](https://en.wikipedia.org/wiki/OpenDocument_technical_specification#Document_representation) is my suggestion with [`PrettyPrinting`](https://help.libreoffice.org/Common/XML_File_Formats#XML_file_structure) `true` in LibreOffice. LaTeX is hard to me.
- I upgraded to latest font and changed your suggestions.
- `ccmp` is from [original Libertine documentation](http://libertine-fonts.org/) (Contextual Chaining Substitution). I'm a simple end user. I'm not an OpenType expert. I'm just learning how to use OpenType in LibreOffice.

khaledhosny:
Thanks!

KrasnayaPloshchad:
>5. I can't reproduce CO₄³⁻ (see last rows in libertinus-sample.odt) with OpenType features...
>
> I dropped the kerning here because it was non-standard, superscript and subscript modifier letters are phonetic symbols and not to be used in math and chemical formulas, these need specialized layout engines.

That was a work around for many layout engines, for example, Micrsoft Word doesn’t provide the way to make superscript and subscript placed into one block; HTML and CSS also have no specialized methods to produce that, in Wikipedia the editors have to procuce by means of vertical-align property.

KrasnayaPloshchad:
I suggest add the kerning into ZWJ to make it still available.

PR #202, opened 2018-07-20
Resizing some math symbols

jco2:
I'm trying to resize some math symbols.
For now I've resized the bar to match the height of the brackets and parentheses as you can see:
resized_bar

Have you any suggestion ?


khaledhosny:
Please cleanup this pull requests, you are adding a whole lot of untracked files which you shouldn’t, and also makes it hard to see the actual change.

khaledhosny:
There are still lots of new changes in the SFD file that do not seem to be related to this change. Also please rebase cleanly on master and squash all the commits.

alerque:
@jco2 With git "starting over" is not the right approach. There are tools to cleanup messy commit history. All this mess can be cleaned up in place, including this PR. You should look into how `git rebase` works, including the edit and squash options.

jco2:
Thanks for the info ! I'm currently looking at this and will see if I can reopen this PR with new commits.

alerque:
When you get your local branch sorted out (rebase it on the latest upstream, squash all the commits, etc.) then you can force-push to the branch this PR was made from and it will update and get rid of all the old stuff.

Issue #63, opened 2017-04-19
Math operators are small


khaledhosny:
The math operators are rather small compared to other math fonts. This might have been OK for the text font, but not so much for the math font.

jco2:
I'm interested in resizing some common math symbols. I'm just new to that so all I need is to edit manually the symbols in FontForge and build the font as explained in the 'readme' ?

ivo-s:
@khaledhosny I am currently looking into this. I managed to get some work done already in my fork: https://github.com/ivo-s/libertinus/tree/math-fix
It's going well, so I decided I would see it through :smiley: 

Because I am currently busy writing my thesis, I am only doing this in a few spare moments. So it might take a bit... If you feel like it, you can check my progress - any feedback is welcome. However, please keep in mind that some glyph references are still broken until I get to them.

The aim is to fix the most obvious issues so that conventionally typeset math looks nice. Because everything is linked, I need to:
* resize/redraw operator glyps so that equations look balanced and comprehensive.
* check references to everything that uses these glyphs and fix it
* because the math axis needs to be united, all forms of brackets need to be moved/resized to be reasonably centred.
* while I'm at it, might as well fix the brackets. The STIX glyps are sometimes too thick and the round brackets () are too round, which sometimes creates too much "air" between the wide part and some symbol after it. Also the bearings need to be optimized, because currently stuff like |0⟩ has asymmetrical spacing.
* then, every vertical and horizontal variant and construction needs to be checked, the coordinates revised and MATH table double checked so that everything fits :smiley: 

I plan to do a pull request after I finish with the operators, which is still a lot of work, but it's the minimal possible step that can be committed.

Some decisions so far:

I decided to use the math axis 260. It is already given in the MATH table, so it is used for constructing fractions in TeX. However, the symbols are centred around a different height, roughly 250. And the guide layer has a math axis at 290. After looking at other math fonts and considering x height, capital height, the need of brackets centering and vertical glyph descents, I decided that 260 would be best.

After studying the letter glyphs, I also decided that all stroke ends should not be straight and rectangular like in STIX, but pleasantly rounded like the Libertinus serifs (that we all love). I think it gives the font a consistent look. On the other hand, the asymmetric endinds in glyphs like + or = look like too much to me.

After everything is done, it should also resolve the issues #164 and #227.

khaledhosny:
Nice work!

[...]

PR #207, opened 2018-08-03
Resize some math glyphs

jco2:
I'm resizing glyphs in the math font (see #63).


khaledhosny:
That last commit is modifies too many lines, please rebase it away. Lets see what glyphs are causing the issue and manually fix them. Automatic fixing almost always breaks things.

To find which glyph is causing the issue, open the generated OTF file, run “Find problems” dialog and select “Intersecting Paths” from the Paths tab.

jco2:
So I've located the glyphs which cause the issue: they have some overlapping glyphs in them such as /uni226F or /uni2271. How can we fix this ?

khaledhosny:
Looks like your new `greater` glyph has the wrong path direction (did you copy the `less` glyph and flipped it?). Element → Correct direction should fix it.

khaledhosny:
Note also that, the mean reason I didn’t do the scaling tell now is that many of these operators are uses as references in other glyphs, and scaling them will break the other glyphs and you will need to check all the glyphs that reference them one by one (I don’t know of any easy way to find all the glyphs that has a given references, you will probably need to script it and assign such glyphs a certain color and review them).

jco2:
Yes, this was the problem. I plan to add more weight to the math operators just like the \in symbol to better match the font.
I wonder how to make a glyph the mirror of the other: like the < glyph can be edited and the glyph > gets the changes too.

jco2:
Thanks ! What do you think of the changes so far ?
I think math symbols shouldn't have rounded corners, especially considering some glyphs get a mixture of both rounded and not rounded corners.

khaledhosny:
Sorry, I’m a bit busy currently, will check it later.

khaledhosny:
I’m split about the rounded corners, part of me wants to keep it as it can be seen as part of the font overall identity, and part of me finds it a bit weird.

alerque:
I'm sorry just an observer here as I rarely use the math font, but can I just say that "font independent" doesn't really make any sense to me? Why purposefully make any part of a font exactly like some other font? Why wouldn't some elements of character come through in math symbols the same way they do in other symbols? I wouldn't expect this at all and indeed would expect some character to come through. I would expect formulas set in _Libertinus Math_ to feel different than ones set in _STIX Math_ or anything else. If you like the glyphs from another font so much why not use that font instead?

jco2:
Actually I expect some non alphabetical math glyphs to look in a certain way leaving little room for variation. This is why I think some non alphabetical math glyphs can fit any font provided they are modified to match the weight and style of the font. That’s what I currently do and I’m satisfied with the results.

khaledhosny:
Sorry for the delay. I generally like the changes, thanks! I think you missed ÷ and few + signs, also the dots on signs like ∓ and ∔. Also signs like ⪯ are now partially resized and look a bit odd.

jco2:
Thank you for reviewing my changes. Unfortunately I'm no longer interested in contributing since I've settled in an other font and I have other priorities. Thanks again for your help with this stuff ! Keep it up this is a really cool project.

khaledhosny:
OK.

PR #228, opened 2018-10-02
Restore design of italic figures and interpolate semibold italic figures

skef:
These changes to the italic and semibold italic figures (as well as uni2074) are explained in the attached PDF. 

[italic_combo.pdf](https://github.com/libertinus-fonts/libertinus/files/2438413/italic_combo.pdf)


skef:
I've created a [repository](https://github.com/skef/libertine_figure_history) with a record of Linux Libertine arabic numeral development that might be a helpful reference for evaluating these proposed glyphs.

khaledhosny:
Thanks for the changes and the detailed write up! Looks good, please rebase and squash all the commits.

skef:
OK, I think it's in shape now

Issue #201, opened 2018-07-19
Not possible to use Libertinus serif semibold in MS Word and other programs


EgUG:
If Libertinus serif bold and serif semibold are installed, it is not possible to use semibold in MS Word 2016, PowerPoint 2016 and Publisher 2016. PowerPoint and Publisher do no not recognize them, Word makes no differences between bold and semibold (see picture).
![capture_001_18072018_104532](https://user-images.githubusercontent.com/41321085/42870849-34bea6aa-8a79-11e8-8a13-27dff869ec2c.jpg)
In Gimp it is no problem to choose the different fonts.

alerque:
This might be a limitation of Word itself, not of the font. Some programs don't actually try to present what is available to the user, they just look for if there is X or not. In this case Gimp might be scanning the available font variants asking "what weights are there is this family" and displaying everything it finds, while Word might be just asking the system "is there a bold version of this family" and displaying whatever the closest approximation of that is.

Do you have examples of other fonts with more than 2 weights showing up properly in your version of Word?

khaledhosny:
Word and Windows GDI can’t handle more than the 4 badic font styles;
regular, bold, italic, and bold italic. There is however a work around for
this, I’ll check if we do this already or not.

>

alerque:
@EgUG Some font foundries work around these limitations by generating distributing their fonts in different groups with the different weights or styles listed under different font names instead of actually being different weights and styles of the same font. This makes them available to older (or just poorly developed) software, but that comes at the expense of making them obnoxious to use in software that _does_ support OTF features and a full spectrum of weights.

As an open-source project development on this font has mostly concentrated on a clean implementation using Unicode standards and standardized open-type features. This makes them quite pleasant to use in modern software. Even if the upstream project does not opt to provide this particular work around, you should know that you can always build it yourself with alternative names for different styles.

This is something that comes up in the context of not being able to reach the open-type features like ligatures and alternative glyphs in some software. There are tools for generating static font files under new names with whatever more advanced features you want enabled or disabled by default for use in programs that otherwise don't have the chops to access those features.

khaledhosny:
The solution here is to  use name IDs 1 (Font Family) and 2 (Font Subfamily) only for the basic 4 variants, and name IDs 16 (Typographic Family) and 17 (Typographic Subfamily) for unrestricted family names.

[...]

PR #229, opened 2018-10-04
Organize display of unencoded glyphs

skef:
I'm "presenting" this as a pull request but I really mean it as the start of a conversation. (Also, these files also include the italic figure changes in my push from two days ago but I can make the two independent easily enough.) 

These deltas add a script ff_sfd_pseudoenc to tools and integrate it into the Makefile's normalization process. This is a script that I wrote (under FontForge modified BSD license, text included in the script itself) to organize the display of unencoded glyphs. Documentation is at:

https://github.com/skef/fontforge_coverage_tools/

but as I've set things up the important thing to remember is to either keep your main FontForge window eight glyphs wide or set `glyph_align` in a `[top]` section of the configuration files to the desired number. 

Organizing the unencoded glyphs has a number of advantages, not least lowering the subjective resistance to adding new unencoded glyphs because of the mess they would make. Another advantage is evident in how I have organized the math font: The `[ssty]` section already reveals some inconsistencies in (at a minimum) the naming and appearance of U1D41A.ssty through U1D433.ssty and U1D5BA.ssty through U1D5D3.ssty. Also the nine variant at U1D7D7.ssty seems to be missing, and I'm guessing uni210E.ssty (in the .ssty "overflow" section) should be named U1D455.ssty. 

And of course organizing the glyphs this or any other sane way makes a given glyph much easier to find. 

So I am making this push request to let you know that this tools is now available and to offer to tailor the configuration files as you would like them. It would probably also make sense to change some glyph names -- the uniE??? series stand out in that those names don't say anything about the contents, especially because they're not actually encoded in the private use area anymore. But that sort of thing can easily wait.


skef:
I've made this a single commit off my italic push request.

Note that although the otf files differ in the binary sense, sfddiff reports no differences when run on the old and new otfs.

Also, something not documented in the putback message: This update also changes the family names of the Serif-Semibolds to "Libertinus Serif" from "Libertinus Serif Semibold", to be consistent with the other serif sfd files.

skef:
Hmm. Well, I haven't been able to trick Travis CI into being happy about the added python 3 script, so this isn't passing the tests. It's a pretty clean script, though. I'll wait for advice on your end.

khaledhosny:
Sorry for not responding so far, but I need more time to study the changes and the code here. The odd semobold font name is intentional, see #201.

tools/ff_sfd_pseudoenc
```diff
@@ -0,0 +1,1153 @@
...
+20000..2A6DF; CJK Unified Ideographs Extension B
+2A700..2B73F; CJK Unified Ideographs Extension C
+2B740..2B81F; CJK Unified Ideographs Extension D
+2B820..2CEAF; CJK Unified Ideographs Extension E
+2CEB0..2EBEF; CJK Unified Ideographs Extension F
+2F800..2FA1F; CJK Compatibility Ideographs Supplement
+E0000..E007F; Tags
+E0100..E01EF; Variation Selectors Supplement
+F0000..FFFFF; Supplementary Private Use Area-A
+100000..10FFFF; Supplementary Private Use Area-B"""
```
khaledhosny: `fontTools.unicodedata` should have some code to help here instead of embedding the Unicode blocks data.

sources/Libertinus_Keyboard_pse.ini
```diff
@@ -0,0 +1,17 @@
+[engkeys]
```
khaledhosny: I don’t like all these configuration files, they would be a maintenance nightmare. Would they are for? Can we do without them  or reduce them to one file or one per family?

tools/ff_sfd_pseudoenc
skef: The tension here is between fewer lines in the script and a dependency. Right now the script's only dependency is on Python 3 and its standard library. That has certain advantages for less sophisticated/integrated projects.

khaledhosny: We depend on FontTools already for other scripts, so I’d rather use it than maintain duplicated code that needs to be updated with future Unicode releases. The less code I have to maintain, the simpler my life is.

sources/Libertinus_Keyboard_pse.ini
skef: It was my expectation in designing the tool that it would be used with a per-family configuration file. And in fact that's how it is set up right now:

    $ grep FamilyName *.sfd
    LibertinusKeyboard-Regular.sfd:FamilyName: Libertinus Keyboard
    LibertinusMath-Regular.sfd:FamilyName: Libertinus Math
    LibertinusMono-Regular.sfd:FamilyName: Libertinus Mono
    LibertinusSans-Bold.sfd:FamilyName: Libertinus Sans
    LibertinusSans-Italic.sfd:FamilyName: Libertinus Sans
    LibertinusSans-Regular.sfd:FamilyName: Libertinus Sans
    LibertinusSerif-BoldItalic.sfd:FamilyName: Libertinus Serif
    LibertinusSerif-Bold.sfd:FamilyName: Libertinus Serif
    LibertinusSerifDisplay-Regular.sfd:FamilyName: Libertinus Serif Display
    LibertinusSerifInitials-Regular.sfd:FamilyName: Libertinus Serif Initials
    LibertinusSerif-Italic.sfd:FamilyName: Libertinus Serif
    LibertinusSerif-Regular.sfd:FamilyName: Libertinus Serif
    LibertinusSerif-SemiboldItalic.sfd:FamilyName: Libertinus Serif Semibold
    LibertinusSerif-Semibold.sfd:FamilyName: Libertinus Serif Semibold

    $ ls *.ini
    Libertinus_Keyboard_pse.ini Libertinus_Mono_pse.ini  
    Libertinus_Serif_Display_pse.ini Libertinus_Serif_pse.ini
    Libertinus_Math_pse.ini Libertinus_Sans_pse.ini
    Libertinus_Serif_Initials_pse.ini

The problem is that what constitutes a subjective family and what constitutes a family at the semantic level can differ.  However, it's also quite easy to set the tool up to look at a single configuration file, or to have some `.sfd`s have family-specific config files and others use the default. [Discussion continued at root-level.]

tools/ff_sfd_pseudoenc
skef: In that case, would you prefer to remove the code from the repository entirely and add a dependency on the script? I already have its repository set up as a python package ( https://github.com/skef/fontforge_coverage_tools) and it's easy enough to make it available via PyPI. (I was holding off on that until I had a bit more experience with it.) 

I'm not envisioning this as something maintained by you (as the Libertinus maintainer) or any other individual project, but that would be maintained via pull requests to `fontforge_coverage_tools`. I just figured it would be easier on everyone to "distribute" the script via font repos rather than require individual contributors to add a new python package on their local machines.

skef: I'm now realizing I was never explicit about this:

1. Although I wrote this tool based on my experience working with Libertinus, I designed as a general-purpose script that can be used by any FontForge-based project.
2. It's `home` is the `fontforge_coverage_tools` repository, and any changes should be made via push request there.
3. It may eventually be used primarily as a locally-installed python package, but I would prefer to get more experience with it before "advertising" it that way.
4. For a variety of reasons (including unsophisticated contributors) many projects may want to use it in the way it is set up in Libertinus now, as an independent script that can copied from `fontforge_coverage_tools` as needed.
5. Projects distributing the script in this way benefit from it having no dependencies, which is why I took the block-quote approach rather adding a dependency on another package.

PR #231, opened 2018-10-11
Extend caps lining figures to (almost) all Serif faces and proportional and tabular spacing

skef:
These changes are explained in this pdf: 
[cap_figs_doc.pdf](https://github.com/libertinus-fonts/libertinus/files/2470155/cap_figs_doc.pdf)

Note that this is a single-commit push based off the current repository rather than my other push.


khaledhosny:
Thanks!

PR #233, opened 2018-10-18
Slight increase to italic proportional figure spacing + tweaks

skef:
This is a more modest pull request than the last few. It does two things:

1. Slightly increase the proportional spacing of figures in Italic and Semibold Italic, which in retrospect I decided was too tight in the previous push.
2. Fixes some little details in the full-size and caps figures in all the standard Serif faces other than Regular and Display-Regular (which didn't have any issues that I could see).

It's hard to summarize all the changes, but they mostly fall into these categories:

1. Removing curve points that do nothing or almost nothing.
2. Removing curve points that do somewhat more but the glyph looks better without them.
3. Correcting the "type" of points:
    1. Making into corners what were incorrectly marked as curves
    2. Marking into tangents what were incorrectly marked as corners.
4. Eliminating some corner-controls for what are really just straight lines (particularly on five). 
5. Restoring the controls on the left bottom bases of one and four, which were missing (or incorrectly interpolated) in some fonts for some reason).
6. Fixing corners in weird places, such as above the counter of six.

With these changes the Serif figures are in pretty good shape other than kerning, which some of the proportional pairs would benefit from. I was going to add some kerning into this update but I decided that was better left for later. (There are probably still issues with base length on one and four, as these seem to be equal in some cases and unequal (presumably for visual balance) in others. But that can wait for better angels than I to stop in.) 

Once again this update is independent of the unencoded glyph conversation/pull request and based off the current repository. 

In case this doesn't go without saying: I'm not in any hurry with these requests. I'm plugging away on a project of sorts here, but my schedule doesn't depend on yours, or for that matter acceptance at any point. (Which is not to say rejection wouldn't be a disappointment ... )


skef:
I'm withdrawing this bundle of changes in favor of integrating them into a more substantial request that will come through soon.

PR #239, opened 2018-11-02
Add download instructions

ChristopherKingChemist:
Added a statement that a zip file exists and gave directions for how to find it.  Letting users know that they must log in to download the file should make this font easier to install.


ChristopherKingChemist:
Sorry; I see it, too.  I'll close and modify this pull request.

PR #240, opened 2018-11-02
Add download instructions

ChristopherKingChemist:
This change tells the user that a zip file exists and where it is located.  That should make it easier for new users to find what they need to install the fonts.  (When I first started using this font, I did not know that a zip file existed.)


khaledhosny:
Thanks!

Issue #191, opened 2018-06-14
Missing super/subscripts - Issues with realscripts package


Doc73:
I would suggest to add, if possible, superscripts and subscripts for Italic and Bold/BoldItalic, to be able to use `realscripts` package.

It would be greatly appreciated.

Many thanks in advance,
Domenico

P. S.: I can add a MWE, if needed.

Doc73:
I did some tests and noticed that this issue is related to `osf` option. Here is a MWE, from which we deduce that the original Linux Libertine O font is not affected by this issue.

```tex
% !TeX program = xelatex
% !TeX encoding = UTF-8
% !TeX spellcheck = it_IT

\documentclass[a4paper]{book}
\usepackage{fontspec}
\setmainfont[Ligatures=TeX,Numbers=OldStyle]{Libertinus Serif}
%\setmainfont[Ligatures=TeX]{Libertinus Serif}
\newfontfamily\libertine{Linux Libertine O}
\usepackage{realscripts}

\begin{document}
\section{Libertinus Serif}
A\textsuperscript{0123456789}

\textit{A\textsuperscript{0123456789}}

\textbf{A\textsuperscript{0123456789}}

\section{Linux Libertine O}
\libertine{
A\textsuperscript{0123456789}

\textit{A\textsuperscript{0123456789}}

\textbf{A\textsuperscript{0123456789}}
	}
\end{document}
```

khaledhosny:
I see. So we don’t have old style superscripts, and Linux Libertine was always using lining figures for superscripts. Would this be an acceptable solution?

Doc73:
Excellent solution for me! I'd prefer lining figures for both super and subscripts.

skef:
`sinf` and `sups` now supersede `onum` in all the main fonts.

Issue #47, opened 2016-10-31
Missing styalistic alternates for Bold


alerque:
In working up #46, I realized there is another issue with the +salt set: none of the special shapes are available in the Bold or Bold Italic variants.

Most notably here the descender on the 'h' is missing:

```TeX
\begin[papersize=a5]{document}
\set[parameter=document.parindent,value=0]
\set[parameter=document.parskip,value=1em]
\font[family=Libertinus Serif,size=16pt]

\font[features="+salt,+ss02"]

\font[]{aâbcçdefgğhiijklmnoöpqrsştuüûvwxyz}\break
\font[style=Italic]{aâbcçdefgğhiijklmnoöpqrsştuüûvwxyz}\break
\font[weight=800]{aâbcçdefgğhiijklmnoöpqrsştuüûvwxyz}\break
\font[style=Italic,weight=800]{aâbcçdefgğhiijklmnoöpqrsştuüûvwxyz}
\end{document}
```

![selection_281](https://cloud.githubusercontent.com/assets/173595/19858822/afbb271e-9f8b-11e6-8d02-ecdef7af331e.png)

khaledhosny:
Do you really need/use that h? I always thought it looked like some IPA/linguistics thing.

alerque:
I do actually use it quite a bit. I'm actually rather shocked I've never seen this before given how much text I typeset with it as a defining feature. However I use it for roughly the same situation across projects and the specific situation is very rarely  bolded—often italicized but rarely bolded.

skef:
Stylistic alternates are now uniform across all the main fonts.

Issue #29, opened 2016-04-04
Localized form for Cyrillic should be expanded


KrasnayaPloshchad:
Here is a test for localized form for Cyrillic:
![1](https://i.imgur.com/tNpHl8N.png)
![2](https://i.imgur.com/KkBnoDm.png)
In this case б should be transformed in all faces, other letters should also has italic variant in Libertinus Serif Italic, Libertinus Serif Bold Italic and so on. BTW I found [Bosnian](https://en.wikipedia.org/wiki/Bosnian_language) sometimes uses Serbian Cyrillic in Bosnia and Herzegovina, is it possible to make it available for Bosnian?

Crissov:
If I understand correctly, the issue is that most southern Slavic writing systems share the same glyph variants, but they are currently only defined in `locl` for certain language codes (e.g. Serbian). Although I agree that this should be fixed at some point, I think it’s also what [`font-language-override`](http://www.w3.org/TR/css-fonts-3/#font-language-override-prop) is intended for in CSS.

KrasnayaPloshchad:
In this case _ѓ_ should also transformed as _ī́_ in all italic faces for southern Slavonics.

khaledhosny:
So the original report is very short in details, and I still don’t know what is being requested here.

I need simple points that can be easily understood by a non-native, e.g. character U+XXXX should use glyph foo for locales X, Y and Z.

Crissov:
I will update [my JSFiddle](https://jsfiddle.net/3se3dpnz) soon.

KrasnayaPloshchad:
Let me explain what happened to me.

The following characters should have their localized glyph for southern Slavics:
U+0431 (б) — should looks like δ in all faces.
U+0433 (г) — should transformed as _ī_ in any italic faces.
U+0453 (ѓ) — should transformed as _ī́_ in any italic faces.
U+0433 (д) — should transformed as _ɡ_ in any italic faces.
U+043F (п) — should transformed as _ū_ in any italic faces.
U+0442 (т) — should transformed as _ɯ̅_ in any italic faces.

When I made the test on wymeditor.github.io, the results to me as follows:
Libertinus Serif: only б, г, д, п, т works in Serbian.
Libertinus Sans: no localized glyph appearing.

khaledhosny:
OK, any samples from fonts doing this right? Is there a definitive list of languages that need this behavior?

moyogo:
@khaledhosny this is a common practice for some foundries.
[...]

PR #249, opened 2018-12-29
Normalize order of per-glyph feature lines (Substitution2, Ligature2 ... )

skef:
The substantive change in this request is to `sfdnormalize.py`. It has been extended to normalize the order of feature-related directives specific to glyphs. Those directives are:

1. Position(2)
2. PairPos(2)
3. LCarets(2)
4. Ligature(2)
5. Substitution(2)
6. MultipleSubs(2)
7. AlternateSubs(2)

The lines will first be put into the relative order above, and within entries in alphanumeric sort order. (The order is taken from `sfd.c` in the FontForge source, but is otherwise arbitrary.) Each .sfd has been normalized with the new script, changing the order of feature entries. 

This normalized order is not all that different from the already imposed ordering of `OtfFeatName` directives. FontForge stores the former in (more or less) the order they were added, and maintains that order when reading them in. But the order itself has no semantic implication. When order is relevant to non-chaining feature subtable entries, it is due to the content of those entries (see end of section 4.e of the Adobe spec: https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#4.e ). (The data for chained rules are stored in the font headers rather than the individual glyphs entries.) 

The (implied) semantic irrelevance of directive order  can also be verified by inspection of `glyphcomp.c:comparelookupsubtable()` in the fontforge source, and by running `sfdcompare` on the Libertinus source file versions before and after normalization with this utility. I have done this and verified that the only "difference" reported relates to the currently empty `calt` table in some of the files. 

I am pushing this change in advance of another that should be coming in the next few days. The idea is to eliminate git differences that just trace to the order in which feature subtables are loaded so that future reviews can focus on substantial changes.

PR #251, opened 2019-01-02
Standardize OpenType Features across faces

skef:
This request involves a large degree of change and will probably take a while to work through the reviewing. So I'm going to document the technical part now and the design part a bit later. 

The technical part relates to the GSUB tables. With these changes, the features of the main font faces (Serif-Regular, Serif-Italic, Serif-Bold, Serif-BoldItalic, Serif-Semibold, Serif-SemiboldItalic,, Sans-Regular, Sans-Italic, Sans-Bold, SerifDisplay-Regular) are now standardized on the model of Serif-Regular. The details and implementation are in sources/features. There are now feature files for all tables other than the experimental combining accents for superiors in Serif-Regular, and the weird Elvin variant in Sans-Bold attached to ss05 (the latter of which I recommend be put "on ice" in a future change). The `README.md` in that directory has more details. 

However, there is something important that document does not mention: `reload_features.py` will not run correctly without some fixes to FontForge, mainly those in fontforge/fontforge#3344 and fontforge/fontforge#3350 . Both of those are merged, so code compiled from the current master will work fine. And anyway, the current `.sfd` files match the contents of the feature file tables in `sources/features`, so it is not necessary to run the script other than to test it. 

The design part of the review involves assessing the glyphs added to allow the feature set to converge across the faces. Well, almost converge: Sans and Serif have somewhat different needs, and Display-Regular does not yet have smallcaps. Other than that, the substitutions are now consistent. 

For your convenience, here is a zip with the output of `sfddiff` comparing the normalized files of this push with those of the current master:  [sfddiff_out.zip](https://github.com/libertinus-fonts/libertinus/files/2720625/sfddiff_out.zip)

Note that there are many glyphs marked as missing in Sans-Regular and Sans-Bold. This was due to a re-encoding snafu that removed a number of "glyphs" that had no outlines or references but only color entries. I can stitch those back in if really necessary, but don't want to bother unless it matters. 

Other than those, here are the explanations for the other *missing* glyphs:

1. Sans-*: zero.slashfitted is now zero.slash.fitted, Q_u is now Q.u (and implemented differently), K.alt and R.alt were swashes copied from the serif fonts
2. Sans-Italic: uniE04F is now t_z (and updated)
3. Serif-BoldItalic: The .superiors were moved into the 1DXX unicode slots. scommaaccent.sc was renamed, uniE02F was renamed W.alt
4. Serif-Regular: uni01E2.sc was renamed
5. Serif-Italic: The .superiors were moved into the 1DXX unicode slots.
6. Serif-SemiboldItalic: uniE04D was renamed
7. Serif-Semibold: The .superiors ..., tcedilla.sc was renamed.

Beyond this, some glyphs have been changed and many have been added. I'll try to provide a better explanation of the changes (or at least an easier route to reviewing them) soon. For now, the main thing to know is that the superiors and inferiors added to Serif-SemiboldItalic are meant mostly as placeholders. The situation with those glyphs outside of Serif-Regular is a mess anyway, and needs an overhaul. Those added or adapted are about as close as those in other faces, such as Serif-BoldItalic. 

If and when this is merged it will resolve or almost resolve the following bugs:

#191 , #47 , #29 

With #29 the glyphs are now present and substituting as in Serif-Italic, but we should nail down the language/script combinations and determine if we need an alternate for uni0453. 

I suggest starting with a look from the high level, and trying some things out, and then clarifying what more information or changes you would like to see before this can go forward.


skef:
Oh, the `case` lookup has been reorganized as I mentioned and you (broadly) agreed with in #236

skef:
I Corrected the map from /degree /r to /degree /F in liga_sans.fig and rebased. If and when I find issues I'll continue to correct, rebase, force-push to my branch, and not the change here unless you advise some other process.

skef:
I fixed a few missing references in Serif-BoldItalic, resynced with the master, and repushed. 

Here is the promised glyph review helper, in the form of a zip file containing an `index.html` to read and individual SVGs of the modified glyphs: [glyph_review.zip](https://github.com/libertinus-fonts/libertinus/files/2729454/glyph_review.zip)

skef:
Any feedback on this in terms of what it would take to get some of the changes integrated? Would it be better to break the content down in a different way?

khaledhosny:
Sorry for not responding earlier. But the changes are big and I need time studying and verifying them.

One aspect I don’t like (without having checked the changes) is having to separately run a script to apply the feature files to the SFD; I’d like to have a single source of truth and features files would be my preference for GSUB features, so if we move stuff there it should be gone from SFD files and the features files are applied at build time.

I understand this requires and new, unreleased FontForge, so that is a big limitation. One alternative would to switch to FontTools’ feaLib for compiling the feature files (like I do in my Amiri font), but this is likely to be tricky with the SFD mix and requires more thought and work.

skef:
A few thoughts on the "authority" issue, none conclusive:

1. In my mind the most relevant observation is that the distance (in time and effort) from the "draft" in this request and any of the end-result you describe is small. Because feature files support `include` and tables defined outside of lookups, one could right now define "top-level" feature files for each of the fonts quite easily. And for information-sharing purposes having the tables that differ in separate files like they are now make sense. So the issues are important but the differences in implementation are small. 

2. I agree that the feature files should be the "authority" going forward. The question is how to achieve that. 

3. At least at this point I would argue against stripping the GSUB content from the files. I know the Metrics window isn't everyone's cup of tea but I've found it quite useful for testing how different combination of features work together. I feel that taking those capabilities away from potential contributors would be a net negative. 

4. If that's right, the problem becomes how to make the feature files authoritative while still retaining the GSUB content. But this is easy and "make check" already provides a model: a hook can be added to the Makefile to reload all the tables (however) and report any differences as errors. In fact the only thing that argues against adding this to `make check` itself is that for a while it will require a very recent version of fontforge to run, so it's best not to require it on the "developer side", while making it available as an option. 

5. If something like 4 is the way to go, then I am personally agnostic about a choice between my script and using "top level" feature files. It's just as easy to strip out all GSUB content and then merge a whole file (as far as I know -- I haven't tried). It would have the advantage of making the files also authoritative for scripts and languages. The disadvantage would be limitation in the table names, but that doesn't seem like a significant problem. I'm therefore happy to rewrite things to work that way if that's your preference.  

tl;dr: I favor making the feature files authoritative while keeping the GSUB content on the model of "make check".

skef:
Oh, I'll add: I wrote the code in #249 because I anticipated something like this model. If the GSUB content remains but the feature files are routinely reloaded, it helps to have a canonical ordering of feature content to limit the diff to substantial changes.

khaledhosny:
Closing in favor of #258.

PR #256, opened 2019-01-25
Fix #252: Libertinus Math – fixed bearings of italic glyphs and kerning

### LINKED ISSUE #252: Bad left-side bearings of mathematical italic glyphs
First, let me thank you for your work on the Libertinus typeface, it looks great!

Originally, I noticed some odd kerning when using Libertinus Math in LuaLaTeX / XeLaTeX. After some research, I think that the fault is with font bearings. Sometimes, they are very asymmetrical. The case where it really stands out are the brackets. This is a result of typesetting using the libertinus package in XeLaTeX:

libmath

The bearings seem to originate from Libertinus Italic, but in math typesetting I think they should be more optimized for the use with upright glyphs. In Latin Modern Math, the bearings are more symmetrical (assuming italic correction).

lsb_comparison

I did some quick fiddling with the bearings in FontForge to make them more symmetrical and tested it:

libmath_2

While I can't say that I am sure that all the characters look fine in this particular context, letters like C, G, W, T, b or p look much better to me.

I am submitting an issue instead of a pull request for several reasons. First, I need to check that I am not wrong :-) I am no typographer and I have no experience with fonts. This means that I am also not sure what the correct bearings should be. They should probably not be precisely symmetrical, especially if there is more "meat" on one side of a glyph...? Another thing is that adjusting the left-side bearing means that some other entries need to be recalculated accordingly. I only know about top accent position and kerning lookup table, but is that all?

So, provided that the bearings should be fixed, it will take a bit of work. All the italic glyphs in the Unicode block 0x1D400 - 0x1D7FF should be checked, because I found asymmetrical bearings in multiple families (serif, sans, bold...). Then, a Python script should recalculate all the relevant parameters. If you don't have the time, I can try to help, although I would need advice on the issues above. Thanks!

ivo-s:
# List of changes

**Glyphs fixed:** u1d434 - u1d49b (italic and bold italic serif latin),  u1d608 - u1d66f (italic and bold italic sans latin), u1d6a4 - u1d6a5 (dotless i and j), u1d6e2 - u1d755 (italic and bold italic serif greek)

**Kerning class lookups changed/fixed:** 'kern' Italic Kerning (both Latin and Greek), 'kern' Sans Italic Kerning and 'kern' Bold Italic Latin kerning (this one replaced 'kern' Horizontales Kerning in Latin lookup 1)

# Workflow justification
This follows the issue #252. First, I tried just editing left-side bearings by hand. Then I realized that I was going to have a lot of fun with kerning tables due to the inconsistency of my changes, which would break kerning classes. I also found out that the kerning classes for the math font are not entirely consistent. I noticed that the vast majority of the right-side space of each glyph (right-side bearing + italic correction) is 20 units wide. I also examined the glyphs in Latin Modern Math and they look approximately symmetrical. For all these reasons, I decided to make my life easier and make all side-spaces symmetrical using a script.

The primary use of the glyphs is as individual characters in bare math mode, where only italic correction is applied and the kerning must look nicely relative to upright glyphs like brackets and operators. If one uses \mathit in LuaLaTeX, then kerning tables would take over, but this seems of secondary importance. So I decided to prioritize symmetrical upright glyph kerning.

# Workflow
The scripts I used are available for reference with comments:
[libertinus_scripts.zip](https://github.com/libertinus-fonts/libertinus/files/2794136/libertinus_scripts.zip)
First, I made all the glyphs symmetrical so that LSB = 20 and RSB + IC = 20 approximately, where IC remains unchanged (uniform_20.py).

Then, for each alphabet variant I generated a table of desired kernings. I took the corresponding glyphs from normal text fonts (LibertinusSerif-Italic etc.) and evaluated inter-glyph distances. For example, distance between AB is A.RSB + B.LSB + kerning. Then I calculated the necessary kerning table for all character combinations and exported it (big_kerning_table_class.py, for greek and non-class kerning pairs I used similar scripts).

I needed to evaluate and group the glyphs that had similar kernings in order to create kerning classes. I did this bit in Mathematica because of time efficiency, but it could be programmed in Python as well. The criteria of glyphs A and B to be in the same kerning class (let's say left-sided class) was the following: for each character X, the kernings AX and BX differ by a value diff(X). If diff(X) is 2 units at most for every X AND identical for every X, then the glyphs are considered to have approximately identical kernings and are grouped together. I plotted the amount of kerning classes, which should be as low as possible, relative to difference tolerance, which should be also as low as possible, and decided on a reasonable compromise. 2 units (2/1000 of an em) should be negligible and already creates a reasonable reduction in the number of classes. I realize that the kerning tables are bigger than before, but on the other hand, they should be faithful to text kerning :-)

The resulting kerning classes and tables are defined in newKerningClasses.py. Finally, a script modifies the classes (kerning_adjust.py).

## Preview and whew...
I generated a comparison in LuaLaTeX. At first, some kerning is demonstrated.
The first line is `\textit{text}`, while the second line is `$\mathit{text}$`. You can see that the kerning is identical, yay! The following section compares the original Libertinus Math to the fixed version.
[test.pdf](https://github.com/libertinus-fonts/libertinus/files/2794175/test.pdf)

For demonstration, I use straight and ordinary brackets, because the effect can be different. For straight brackets, the visual kerns are determined by vertical strokes at the edge, while for round brackets it might be the top serifs. It's hard to make everything consistent without the option of kerning in math mode.

OK, this took a while and I learned a lot about FontForge and what it actually does, fonts and metrics and how TeX applies them :-) Hopefully the result is better than before. Also my first time using git! :tada: I managed to normalize, make, stage and commit, so everything should be in order. Please review the changes at your convenience and let me know if the result is OK.


khaledhosny:
Thanks, looks good!

I doubt that LuaTeX will use kerning classes in math mode (but I need to check that), XeTeX definitely does not do any kerning in math mode. It does not hurt though, and the side bearing changes are a clear improvement.

PR #257, opened 2019-01-28
Fix #242: vertical variants of math symbols

### LINKED ISSUE #242: \langle and \rangle automatic sizing (bad behaviour)
The vertical bar is not properly sized when used with angled brackets. I have tested the same code with other fonts and the discrepancy does not appear. The following picture is the result obtained by this line: `\[\left|a\right\rangle |a\rangle \left|\sum a\right\rangle\]`
screenshot from 2018-11-02 19-13-26
The only math packages used were `mathtools` and `unicode-math`.

ivo-s:
Incidentally, I encountered the same problem as #242. Happened only with XeTeX, not with LuaTeX. The problem was that the basic sizes were missing in the MATH table for certain glyphs (see diff).


khaledhosny:
BTW, you need to put `Fixes #242` either in the commit message or in the pull request description to automatically close the issue when the pull request is merged (https://help.github.com/articles/closing-issues-using-keywords/).

ivo-s:
Sorry, I'm a bit a of a newbie; it took much hair-pulling before I figured out how to make my fork only one commit ahead of upstream. I read some git help pages, but I forgot about that feature.

PR #258, opened 2019-01-30
Top-level feature files

skef:
Since I said it wouldn't be much work, and your preference is for full feature files, I decided to put an alternate request together along those lines. This isn't fully polished yet but can serve as a reference for discussion. 

Compared with the other request this solves two additional problems:

1. It constructs a feature file for each of the "main" font files from sources in `sources/features`.
2. It adds support for replacing all GSUB data with the contents specified in that feature file. 

The first of these was the bulk of the work. Because there are at least five variants (combinations of italic/non-italic, display/non-display, and sans/non-sans) I decided a top-level file for each variant would involve too much duplication. That means some sort of macro language is needed, and I chose the one now mentioned in the top-level README, which is a python package. The result is a very straightforward  `ifdef` and `include`-based specification of the features. There are a few very small changes to the lookup contents corresponding to bugs. The table names were changed to correspond to their existing names while being feature-file compatible. 

(I hope you appreciate the political risk I am subjecting myself to, as an American, by not just rewording everything in English without asking anyone.)

The other significant changes are the new script in `tools` to do the gsub stripping and reloading, and `Makefile` work. What were called `normalize` and `check` are now `normalize-lite` and `check-lite` respectively. Those build targets now also reload the GSUB content and verify the GSUB content has not changed. I could have added your underline hook in `build` to this process, but as it also adds glyphs I decided to defer that. (I removed the language system include as it wasn't needed and I had deleted that file.) 

The easiest way of checking this work is to compare the generated `.otf` files with those of the other pull request. The output of `sfddiff` for each pair is in this archive: [diff.zip](https://github.com/libertinus-fonts/libertinus/files/2812731/diff.zip)

Most of the changes relate to the underline support. The other changes are:

1. The Elvish support is no longer in Sans-Bold. (If we go this direction I will archive a file with that map for reference, but I think it's better to leave it out of the generated fonts.
2.  `Q` and `q.sc` now have salt entries in the Sans fonts. 
3. There are some additional glyphs in Serif-Semiboldtalic (I must have added this a little later than the comparison point.)
4. SerifDisplay-Regular had a few tables that shouldn't have been there but I missed. One is a second liga table and the other is a c2sc that duplicated `case`. Both of these are gone in this push. 

Even if the strip/reload solution for GSUB is not what you were looking for, the feature file specification should be of interest.


khaledhosny:
English is good, I don’t know any German so all these existing lookup names mean nothing to me. You need to install `preprocess` on Travis to fix the build. My preference is to strip the SFD files from any GSUB lookups and merge the features files at build time, right before generating the OTF files. GPOS lookups should stay in the SFD files as that is easier for editing them.

skef:
I've force-pushed a new version with the requested changes. Note that the travis build now "works" in the sense that it uses `preprocess` and doesn't report an error, but doesn't *work* because the version of fontforge it uses doesn't have the feature-file fixes (either mine from late last year or your more recent completeness work). 

I verified with `sfddiff` that the generated `.otf` files are the same as in the earlier pushes and the earlier (and presumably dead) push request.

`make normalize` (and `check`) now strips any GSUB tables from the "main" fonts (those with feature files). That way an attempt to add substitutions through the `.sfd`s rather than the feature files should be caught early. 

Lookup table names are now in English. Some names are redundant (although not more than before). I tried to make the names informative where that could be helpful. 

With this "model" it isn't clear how a contributor should go about testing substitution changes. Kerning tests are also (arguably?) more annoying, in that rather than typing text into the FontForge metrics window and applying features, one has to track down the glyph names to view the spacing. *I* don't care about this, because I can just type a few commands and make myself a temporary `.sfd`. But other contributors could benefit from some suggestions in the documentation (and, if there's a significantly better and still easy way, I would benefit too). 

(I inadvertently checked in rebuilt docs, but my versions seem fine. I also added a `cleaner` target to the `Makefile` and moved files under version control to it, because it seemed silly to be deleting those while cleaning up temporary files like `.lnt`, `.chk` and the generated feature files.)

skef:
Oh, I also got the old and new maintainers of `preprocess` in touch with each other, and the latter now have `PyPI` permissions. That should be sorted out shortly, after which `pypi install preprocess` should work and the travis hook and `README.md` can be updated.

Makefile
```diff
@@ -4,66 +4,108 @@ VERSION=6.7
...
 
+FEADEFS=
+
 export SOURCE_DATE_EPOCH ?= 0
 
 all: otf
 
 otf: $(OTF)
 doc: $(PDF) $(OPDF)
+feature-files: $(FEA)
```
khaledhosny: I don’t think we need this target.

Makefile
```diff
@@ -121,4 +163,7 @@ dist: check $(OTF) $(PDF) $(OPDF)
 	@zip -rq $(DIST).zip $(DIST)
 
 clean:
-	@rm -rf $(DIST) $(DIST).zip $(CHK) $(MIS) $(DUP) $(NRM) $(OTF) $(PDF)
+	@rm -rf $(DIST) $(DIST).zip $(CHK) $(MIS) $(DUP) $(FEA) $(NRM) $(LNT)
+
+cleaner: clean
```
khaledhosny: `make clean` should remove any generated file, no need for another target.

sources/features/gsub.fea
```diff
@@ -0,0 +1,752 @@
...
+# #endif
+} locl;
+
+feature frac {
+
+ script DFLT;
+     language dflt ;
+      lookup frac_contextual;
+      lookup frac_unicodepoint;
+
```
khaledhosny: The lookups are all the same for all language systems, so you can replace this feature with (but please make sure this produces the same binary, there might be differences I missed):
```
feature frac {
  lookup frac_contextual;
  lookup frac_unicodepoint;
} frac;
```

sources/features/gsub.fea
```diff
@@ -0,0 +1,752 @@
...
+      lookup frac_unicodepoint;
+     language CRT  exclude_dflt;
+      lookup frac_contextual;
+      lookup frac_unicodepoint;
+     language TRK  exclude_dflt;
+      lookup frac_contextual;
+      lookup frac_unicodepoint;
+} frac;
+
+feature sups {
```
khaledhosny: Same here and many other features below.

sources/features/gsub.fea
```diff
@@ -0,0 +1,752 @@
...
+      lookup ccmp_contextual;
+     language CRT  exclude_dflt;
+      lookup ccmp_contextual;
+     language TRK  exclude_dflt;
+      lookup ccmp_contextual;
+} ccmp;
+
+feature ss01 {
+  featureNames {
+    name 3 1 0x409 "Low diaeresis on 'A', 'E', 'O'";
```
khaledhosny: The shorter `name "Low diaeresis on 'A', 'E', 'O'";` should work.

sources/features/liga.fea
```diff
@@ -0,0 +1,41 @@
...
+  sub c uni200D k by c_k;
+  sub t uni200D z by t_z;
+} liga_standard_ligatures;
+
+lookup locl_undo_fi {
+  sub f_i by f i;
+  sub f_f_i by f_f i;
+  sub longs_i by longs i;
+  sub longs_longs_i by longs_longs i;
+} locl_undo_fi;
```
khaledhosny: I prefer to have the feature block to be also in these files instead of `gsub.fea`, unless there is a reason for not doing it.

Makefile
skef: It's so that if you're editing the feature files and want to verify your changes you don't have to wait through generating otfs just to see the files as a side-effect.

Makefile
skef: I don't get this one. `.gitignore` and the genuinely temporary files work well together, in that the latter are ignored for purposes of diffing and committing. But with the current Makefile if you do a `make clean` and then `git commit -a`, you delete a bunch of files intended to be under version control. Why would you want that? Isn't there an important difference between a generated file intended to be checked in when it changes, and a generated file only ever meant to be temporary? 

I mean, I'll do it. I say what I think and I do what I'm told™. But I don't get it.

sources/features/gsub.fea
skef: (will be) fixed.

sources/features/gsub.fea
skef: I fixed the others. This one does actually have an exception in Hebrew. It comes from the source files and I don't know enough about the potential issues to change it unilaterally.

sources/features/gsub.fea
skef: It did work.

sources/features/gsub.fea
```diff
@@ -0,0 +1,752 @@
...
+feature locl {
+
+ script cyrl;
+     language SRB  exclude_dflt;
+      lookup locl_cyrillic;
+
+ script latn;
+     language AZE  exclude_dflt;
+# #ifndef SANS
+      lookup locl_undo_fi;
```
skef: It turns out these locl_undo_fi lines "work" only because FontForge is clever enough to implicitly "move" them below the liga entry, presumably because of the characters they reference. That's nice of FontForge but it's better to have them where they need to be, so I'm moving them into a separate feature.

sources/features/gsub.fea
khaledhosny: I saw it but see no reason for it to be intentional, definitely a typo.

Makefile
khaledhosny: The idea is that if I `make clean` then `make` then any generated file should be definitely generated. Ideally OTF and PDF files shouldn’t be under version control to begin with, but since font building pipeline is so fragile I keep them under version control for when I need to debug a regression and find whether it is a font or tool regression.

Makefile
khaledhosny: Still not sure what kind of changes one would like to see, but it doesn’t hurt either.

sources/features/liga.fea
skef: I went back and forth on this and convinced myself that this is the better overall model. I think I also prefer it for "general understanding of what's going on", but there is also a more mundane technical reason: Lookup tables can refer to one another, as the contextual tables refer to the superior and inferior tables. To make all tools happiest you want to list included tables before including tables, which imposes a partial ordering. However, that ordering is not necessarily compatible with the GSUB feature application ordering, hence the separation.

Makefile
skef: As someone who fiddles around with Libertinus I want a target that leaves me with exactly the set of files under version control. 

What if I make it "less-clean" and "clean" and no one has to know about "less-clean" who doesn't peek?

sources/features/gsub.fea
skef: Removed in next push

skef:
Alright, I just pushed the frac and clean tag swap changes. 

Some bad news, which actually may be irrelevant given your plans: I have been relying on `sfddiff` when checking these changes in relation to the older table-merge version (in the just-closed push request). That test isn't sufficient, and the generated `.otf` files don't work. 

It's possible I have just been mistaken about feature file ordering. My impression was that the order of application was determined by `feature` ordering. FontForge is acting like it should be determined by `lookup` table ordering. (It's inserting features in lookup table order.) So before this moves forward we should verify and fix the generated feature files by testing with some other tool. And I guess that's fine because that's what you planned to do anyway. 

(The only real challenge here is verifying how the file should be structured. Actually generating that file from sources similar the present ones should be easy. All the individual tables have the right entries, and the `#ifdef` logic is also correct.)

Should I play around with feaLib and see what it does?

skef:
I've thought about this more, and I'm pretty sure I can reorganize the content so that regardless of whether FontForge is doing the right thing in general, the result will be correct. I'm going to go exercise and try that out, and with luck I'll be able to do a corrected push 3-4 hours from now. 

You'll likely get the feature blocks in the included files as a side effect, given that FontForge is, in effect, forcing a unified lookup/feature ordering.

khaledhosny:
> Some bad news, which actually may be irrelevant given your plans: I have been relying on `sfddiff` when checking these changes in relation to the older table-merge version (in the just-closed push request). That test isn't sufficient, and the generated `.otf` files don't work.
> 
> It's possible I have just been mistaken about feature file ordering. My impression was that the order of application was determined by `feature` ordering. FontForge is acting like it should be determined by `lookup` table ordering. (It's inserting features in lookup table order.) So before this moves forward we should verify and fix the generated feature files by testing with some other tool. And I guess that's fine because that's what you planned to do anyway.

The significant order when OpenType layout is applied is the lookup order; the lookups are collected based on active features and are then executed in their order in the font (the reality is a bit more complex since the lookups of certain features, e.g. `ccmp`, are always applied first by some layout engines).

I think feature file compilers will insert lookups in the order they are specified in the feature file.

> (The only real challenge here is verifying how the file should be structured. Actually generating that file from sources similar the present ones should be easy. All the individual tables have the right entries, and the `#ifdef` logic is also correct.)
> 
> Should I play around with feaLib and see what it does?

If it is not much work, yes, otherwise I don’t want to block this PR on this.

Makefile
khaledhosny: That would be `git clean -fdx`, no need for a make target :)

sources/features/gsub.fea
skef: This was FontForge's ordering problem. Worked around.

skef:
The push of a moment ago works around the FontForge ordering problems. As I suspected it would, it also moves the feature definitions into the included files. I did a bunch of testing with these, as well as regenerating the sample files (after copying Serif-Regular where it would be found), and the substitutions all seem to be working as expected with these `.otf`s. 

Note that the ordering problem with `frac_contextual` and the `sinf/sups` tables required duplicating the latter using different names. Hence the extra directives in `small.fea`. 

Anyway, I think this is in good shape now. I believe the generated feature files will also work correctly with other (well-implemented) tools.

sources/features/calt.fea
```diff
@@ -0,0 +1,26 @@
...
+  sub Q' lookup chaining_substitutions [u v];
+# #else
+  sub Q' lookup chaining_substitutions [u u.sc v v.sc];
+  sub q.sc' lookup chaining_substitutions [u.sc v.sc];
+# #endif
+} calt_contextual;
+
+feature calt {
+      lookup calt_contextual;
+} calt;
```
khaledhosny: In features that has one lookup and the lookup is not referenced any where else, you can even use the shorter syntax:
```
feature calt {
  # lookup content
} calt;
```

and skip the named lookup, the end result will be the same.

skef: I did know this - I prefer the separate tables. Do you want me to edit them down?

khaledhosny: No, I can do this later.

skef:
> The significant order when OpenType layout is applied is the lookup order; the lookups are collected based on active features and are then executed in their order in the font (the reality is a bit more complex since the lookups of certain features, e.g. `ccmp`, are always applied first by some layout engines).

So feature order is actually irrelevant then, expect when defining lookups within features? Then I was definitely mistaken before. I'll have to go back and look at the feature files FontForge generates and see if they're consistent with that.

khaledhosny:
> > The significant order when OpenType layout is applied is the lookup order; the lookups are collected based on active features and are then executed in their order in the font (the reality is a bit more complex since the lookups of certain features, e.g. `ccmp`, are always applied first by some layout engines).
> 
> So feature order is actually irrelevant then, expect when defining lookups within features?

Yes, I believe so.

khaledhosny:
Please merge (I prefer “Rebase and merge” as I don’t like merge commit, but do whatever you find best).

skef:
The weird thing about that is that you can reference a lookup in another lookup, like the old font was doing with the contextual frac table and the sinf/sups tables. If the lookup order is primary, shouldn't you be able to reference a later table? Is FontForge wrong to enforce the definition order rule (it does do that -- I checked)?

skef:
@khaledhosny: I accepted the project invitation (thanks for that) but I don't think I quite have write permission at this point. I don't see the controls.

skef:
I presume something described [here](https://help.github.com/en/articles/managing-team-access-to-an-organization-repository) or [here](https://help.github.com/en/articles/managing-an-individuals-access-to-an-organization-repository) needs fiddling.

khaledhosny:
> The weird thing about that is that you can reference a lookup in another lookup, like the old font was doing with the contextual frac table and the sinf/sups tables. If the lookup order is primary, shouldn't you be able to reference a later table? Is FontForge wrong to enforce the definition order rule (it does do that -- I checked)?

Lookups can reference later lookups, but not in feature files (a limitation of the spec). A clever implementation would defer resolving referenced lookups to the end, but it probably complicates the code for little gain.

khaledhosny:
I changed the default member permission to “write” (it was “read” by default), you should see the controls now.

PR #268, opened 2019-04-03
README: fix indentation of preformatted block

waldyrious:
Should use 4 spaces rather than 3 for correct rendering.


khaledhosny:
Thanks!

PR #274, opened 2019-08-04
Added TURNED CAPITAL F at U+2132

hectormonacci:
There are 3 characters, known as Claudian Letters, that Roman Emperor Claudius intended to add to the Roman alphabet. Two of them are already present in Libertinus Serif (inverted C at U+2183 [Ↄ] and LATIN CAPITAL LETTER HALF H at U+2C75 [Ⱶ]).

I just added the missing one: TURNED CAPITAL F at U+2132. It looks this way: Ⅎ, and it is needed in order to represent old inscriptions or to compose Robert Graves' novel "I, Claudius" (chapter XVII).


khaledhosny:
Thanks, i did a few osmotic changes and merged.

Issue #279, opened 2019-09-05
Placement of combining vertical bar below (U+0329)


kopeckyf:
The combining vertical car below (UTF-8 0329) is shifted when combined with some characters.

When the Latin lower case letter l is combined with the vertical bar below, the resulting bar is shifted a little too much to the left. It should be put in the centre of the l. 

The bug arises only in `LibertinusSerif-Bold.otf` and `LibertinusSerif-Semibold.otf`. I have not seen it when the vertical bar below is combined with any other character, and the problem does not arise in `LibertinusSerif-Regular` nor in italics.

Here's an illustration of the issue:

![grafik](https://user-images.githubusercontent.com/10461729/64342478-df782280-cfea-11e9-9bff-8090d6855c59.png)

For some glyphs that have non-zero depth, the combining vertical bar below collides with the glyph. I've seen this with the Latin lower and upper case ezh, and Latin upper case J. This bug occurs in all styles of the Libertinus family.

Here's an illustration of the issue (j and g have the desired placement):

![grafik](https://user-images.githubusercontent.com/10461729/64342408-b8b9ec00-cfea-11e9-80e7-054cfa333142.png)

I'd like to thank the developers of Libertinus for their help with the placement of the top tie bar, which was a similar issue!

kopeckyf:
@khaledhosny As far as I can see, the issue is related to the anchor point `unten_punkt`, which is not present in Semibold lowercase l and the other glyphs. Would you like me to add the anchor and initiate a pull request for the change? Since I'm rather new to FF, I'd like to leave exporting the .otf files to you.

khaledhosny:
Sure, please send a PR and I’ll have a look. Thanks.

Issue #30, opened 2016-04-04
U+02E5–U+02E9 should composed in sequence for contour tones


KrasnayaPloshchad:
Several fonts such as Arial, Segoe UI, Times New Roman, DejaVu fonts, Charis SIL and Doulos SIL can do [this](https://en.wikipedia.org/wiki/Tone_letter), but Libertinus font cannot.

Crissov:
This has also been requested in a comment to [bug #300 of Linux Libertine on SF.net](https://sourceforge.net/p/linuxlibertine/bugs/300/). The original bug is about U+1DC4–9  from the [Combining Diacritical Marks Supplement](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks_Supplement) block.

khaledhosny:
I’m tempted to close this as WONTFIX, not every font can/should support every character in Unicode and I don’t see what much value support tone letters in Libertinus would added, you need at most one font to support them and they should then work with any font.

Crissov:
Complete and correct IPA coverage would be a nice feature that not many typefaces possess. This issue would be a small part of that goal, which could get a label or milestone of its own (and probably low priority). WONTFIX sounds so alienating and final.

khaledhosny:
Completeness for completeness sake is just waste of resources. That being said, patches are welcome :) (and here it should be as easy as copying the glyphs and OTL rules from another font with a compatible license).

Crissov:
Completeness assures potential users that they won’t run into common problems, even if they’d never need the missing or malformed glyphs. 

Character coverage (letters and punctuation) is not all, though, because sometimes you need adjusted glyphs or custom font features to really claim proper support for a certain writing system or technical notation. I don’t know whether there is an openly available test suite or certification programme for that, but I hope there is. I would like to see a badge that is awarded to all typefaces that properly cover the writing systems of all official languages of the European Union, for instance, or the Pan-Nigerian Alphabet or, to get back on topic, everything frequently needed in linguistics (IPA, UPA, ∅, ⟨⟩ or 〈〉, …) or other sciences – although “mathematics” covers a lot already.

Anyhow, is the SIL license compatible? As noted in the OP, their fonts usually support this particular use case.

moyogo:
The tone letters in DejaVu fonts are public domain since they were drawn from scratch.

khaledhosny:
[...]

PR #283, opened 2019-10-02
unten_punkt anchors for Latin lowercase j, l, and ezh, uppercase J, Ezh

kopeckyf:
This commit aims to solve #279

### DIFF
--- a/sources/LibertinusSerif-Bold.sfd
+++ b/sources/LibertinusSerif-Bold.sfd
@@ -11709,6 +11709,7 @@ Encoding: 439 439 514
 Width: 632
 GlyphClass: 2
 Flags: MW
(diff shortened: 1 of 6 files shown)


khaledhosny:
Why did you added the tone marks, they are not requested in #279 (see #30 for why they were remove in the first place).

khaledhosny:
I fixed and pushed the first commit. Please open one pull request for each kind of change for easier review.

kopeckyf:
Sorry, the two commits got appended in error. Would you accept another pull request for the short equal sign, and would it still be acceptable to you to copy the tonal letters from DejaVu?

khaledhosny:
Yes to both.

PR #296, opened 2020-02-12
Anchor fixes for r with dot below and amacron in Serif-Italic, added raised modifier I, ŋ, ɨ, ɛ, ʉ

kopeckyf:
These commits have hotfixes for two anchoring issues in LibertinusSerif-Italic:

- move the dot in `uni1e5b`, r with dot below, so that the dot is continuous with the r's body
- in `uni0101`, a with macron, add the `above` anchor so that ā can be stacked

The following glyphs are added in (at least some of) LibertinusSerif-Regular, -Italic, -Semibold, -SemiboldItalic, -Bold. They are not added in LibertinusSeminbold-Italic, because the glyphs that they are based on are largely missing there. 

- `uni1d35`, raised capital I
- `uni1d51`, raised small ŋ
- `uni1d4b`, raised small open e
- `uni1da4`, raised small i with stroke
- `uni1db6`, raised small u with bar

------

The files have been normalized and compiled without error. 
Btw, the italic variants show a strange behaviour when opened with FontForge (20190801). I opened an [issue in the fontforge repository](https://github.com/fontforge/fontforge/issues/4098) for this.


khaledhosny:
I see normalization issues, but I fixed them (`make normalize` should have handled this, I don’t know why it doesn’t for yo given it is handled  both on my machine and on Travis).

kopeckyf:
Hi @khaledhosny, thanks for this. Could you give me an idea what went wrong in the normalisation? I did `make normalize` before uploading, but possibly something is wrong with my config. 
Also, would you accept some further changes before the next release, and if so, by when should they be done? Nothing major - just some kerning corrections for some rather rare characters, etc.

khaledhosny:
I ran `make normalize` and there was a few changes, normally there shouldn't, so either something is broken or you didn't commit the changes after running it. 

I can wait for the release until all the changes you want are in.

PR #297, opened 2020-02-13
Generate Unicode glyph coverage chart(s)

### LINKED ISSUE #295: Unicode support overview
It would be nice if there was a readme table to keep track of Unicode coverage (cf. https://en.wikipedia.org/wiki/DejaVu_fonts#Unicode_coverage ), e.g.
```
U+0000 to U+007F Basic Latin                             : a / 128
U+0800 to U+00FF Latin-1                                 : b / 128
U+0100 to U+017F Latin Extended-A                        : c / 128
U+0180 to U+024F Latin Extended-B                        : d / 208
U+0250 to U+02AF IPA Extensions                          : e /  96
U+02B0 to U+02FF Spacing Modifier Letters                : f /  80
U+0300 to U+036F Combining Diacritical Marks             : g / 112
U+0370 to U+03FF Greek and Coptic                        : h / 135
U+0400 to U+04FF Cyrillic                                : i / 256
U+0500 to U+052F Cyrillic Supplement                     : j /  48
…
U+1AB0 to U+1AFF Combining Diacritical Marks Extended    : k /  15
…
U+1C80 to U+1C8F Cyrillic Extended-C                     : l /   9
…
U+1D00 to U+1D7F Phonetic Extensions                     : m / 128
U+1D80 to U+1DBF Phonetic Extensions Supplement          : n /  64
U+1DC0 to U+1DFF Combining Diacritical Marks Supplement  : o /  63
U+1E00 to U+1EFF Latin Extended Additional               : p / 256
U+1F00 to U+1FFF Greek Extended                          : q / 233
…
U+2070 to U+209F Superscripts and Subscripts             : r /  42
U+20A0 to U+20CF Currency Symbols                        : s /  32
U+20D0 to U+20FF Combining Diacritical Marks for Symbols : t /  33
…
U+2C60 to U+2C7F Latin Extended-C                        : u /  32
U+2C80 to U+2CFF Coptic                                  : v / 123
…
U+2DE0 to U+2DFF Cyrillic Extended-A                     : w /  32
…
U+A640 to U+A69F Cyrillic Extended-B                     : x /  96
…
U+A700 to U+A71F Modifier Tone Letters                   : y /  32
U+A720 to U+A7FF Latin Extended-D                        : z / 174
…
U+AB30 to U+AB6F Latin Extended-E                        : α /  56
…
U+FE20 to U+FE2F Combining Half Marks                    : β /  16
…
```

alerque:
Closes #295.

Also mucks about with some other stuff:

* Cleanup other documentation
* Prepare Makefile for easier use in future CI checks

*Work in progress, expect history rewrites as this takes shape!*


documentation/Design-Guidelines.md
```diff
@@ -41,11 +41,12 @@
 
 # Libertinus design guidelines
 
-These are short guidelines for all volunteers who love libre fonts in general, the Libertinus font family especially, and who think about contributing to the latter.
-Fonts contain dozens of glyphs. Libertinus for example has several thousands of glyphs, and they all fit together while being different. The glyphs have to fit together - in the aspect of design, their optical weight, and their spacing. Some technical issues have to be met to achive that.
-Sounds complicated? Nevermind, we will get you into it. This is what these guidelines are for.
+These are short guidelines for all volunteers who love libre fonts in general, the Libertinus font family in specific, and who are considering contributing to the latter.
```
waldyrious: This is just a stylistic choice, but ["in general / in particular"](https://www.google.com/search?q=%22in+general%22+and+%22in+particular%22) seems to be more common than ["in general / in specific"](https://www.google.com/search?q=%22in+general%22+and+%22in+particular%22) (by about an order of magnitude, according to the linked Google search results). Both are valid, but using a more common expression can make this passage slightly more approachable.

alerque: Fair enough. I'm accustomed to "in general", but either way is better than "...especially" — which would have worked with other wording at the beginning but the whole thing was just a miss-match.

Along with fixing the word you brought up I did a little more work moving through that file, but frankly it needs a _lot_ more help!

Makefile
```diff
@@ -14,30 +14,18 @@ BUILD=$(TOOLS)/build.py
...
-       $(NULL)
-
-SFD=$(FONTS:%=$(SOURCEDIR)/$(NAME)%.sfd)
-NRM=$(FONTS:%=$(BUILDDIR)/$(NAME)%.nrm)
-CHK=$(FONTS:%=$(BUILDDIR)/$(NAME)%.chk)
-DUP=$(FONTS:%=$(BUILDDIR)/$(NAME)%.dup)
-LNT=$(FONTS:%=$(BUILDDIR)/$(NAME)%.lnt)
-OTF=$(FONTS:%=$(NAME)%.otf)
+# Discover all font source files
+SFD=$(wildcard $(SOURCEDIR)/*.sfd)
```
khaledhosny: Actually, I prefer listing the font files explicitly. I don’t want some random SFD file that happen to be in the sources directory to be built and added to the release tarball.

Makefile
```diff
@@ -46,7 +46,7 @@ check: $(LNT) $(CHK) $(DUP)
 nofea=$(strip $(foreach f,Initials Keyboard Mono,$(findstring $f,$1)))
 
 $(BUILDDIR)/%.ff.otf: $(SOURCEDIR)/%.sfd $(GSUB) $(BUILD)
-	echo "      BUILD  $(*F)"
+	$(info =>     BUILD  $(*F))
```
khaledhosny: My VIM configuration highlights NBSP and I input it often that is usually accessible from my keyboard layout, so I’d rather use it than these faux arrows.

Makefile
```diff
@@ -132,6 +132,22 @@ $(DOCSDIR)/preview.svg: $(DOCSDIR)/preview.tex $(OTF)
 		$< 1>/dev/null || (cat $(basename $<).log && false)
 	mutool draw -q -r 200 -o $@ $(basename $@).pdf
 
+define unicode_coverage_table =
+	$(shell echo ttx -t cmap -o - $1.otf)
```
khaledhosny: What is the `ttx` call for?

alerque: It's not yet, it's just one alternative for collecting data to actually generate the table with. This is a WIP ;-) I'm actually leaning towards `pyfontaine` instead to collect the same data, but for that I'm kind of waiting for [this upstream bug fix to land](https://github.com/googlefonts/pyfontaine/issues/110).

Makefile
alerque: Roger that.

Makefile
alerque: Okay, I can see the logic in that. It makes it harder to hack on, but also harder to break. It also has the advantage of being able to manually specify an order, which will be useful elsewhere. I'll re-instate a manual control over that.

khaledhosny:
Travis on master is suing Python 3.6.x, as we no longer depend on FontForge Python module for building the fonts.

alerque:
Ooo, interesting, thanks for the update.

alerque:
By the way if you're wondering what happened to this, work is kind of held up until [this pyfontaine PR is merged](https://github.com/googlefonts/pyfontaine/pull/111). I could hack together various ways around this, but I thought it was a really glaring bug with a simple fix and was hoping it would get merged quickly and I could go on with my plans to use it here. If I can't make any headway there I'll start looking for alternatives.

waldyrious:
@alerque would a PDF like [this](http://users.physik.fu-berlin.de/~tburnus/linlibertine/Overview-Linux-Libertine.pdf) (found [here](http://users.physik.fu-berlin.de/~tburnus/linlibertine/)) be in scope for this PR?

alerque:
@waldyrious Believe it or not it's actually a little more complicated to get a textual representation of what ranges are covered (Unicode 13 is a little more expansive than 5.0 when that was generated) but yes, generating a PDF like that is a thing that can happen based on this work

alerque:
The commits I just pushed sketch out one possible way this could be done. I'm not exactly sure what information would be most useful and in which format. The actual list of potential ranges covered as defined by Unicode 13 is somewhat extensive and useless. I've tried culling the list of possible matches a bit and it looks pretty good for most of the fonts at first blush, but some of the fonts show up many (dozens) more ranges than others and I don't understand why. It's the ones with less coverage too (.e.g. they Keyboard face). A closer look shows lots of weird anomalies.

1. What subset of Unicode Blocks do we want coverage reporting for?
1. What do we want the output format to be? Exactly.

image

khaledhosny:
I have no answer to any of these questions, unfortunately. This wild inconsistency in cover between the fonts is one of the reasons I dropped `fntsample` generated PDFs as they are just too many.

khaledhosny:
May be splut the non-Unicode coverage parts to a separate pull request? I’d like to merge these.

alerque:
Yes I could do that. I've been laid up with strep-throat for a few days but I'll split out the other build related fixes separately.

waldyrious:
Just for completeness' sake, here's a [pdf](https://github.com/alif-type/libertinus/files/4467212/LibertinusSerif-Regular_coverage.pdf) that I generated using [fntsample](https://github.com/eugmes/fntsample) just to see what the output would look like.

PR #307, opened 2020-03-25
Fix U+2639, U+263A and U+263B (frowning & smiling faces)

### LINKED ISSUE #183: Weird shape for U+263A, white smiling face
I noticed this when visiting [the Wiktionary page for this character](https://en.wiktionary.org/wiki/%E2%98%Ba). Look at the discrepancy between the page title (which uses Linux Libertine) and the other instances of the character in that page.

☺, U+263A ("White Smiling Face") was the original smiley face encoded in Unicode. Unfortunately due to various emoji libraries choosing to use it for emojis with various features (such as blushing cheeks, closed eyes, and sometimes even an open-mouth grin), the Unicode consortium had to release a new character specifically for this expression, to ensure consistency: 🙂, U+1F642 ("Slightly Smiling Face").

In any case, the shape used in Libertinus for the original ☺ U+263A (which is also the same as from Linux Libertine) is clearly non-standard and deviating from the original intent. Here is the glyph compared to Liberation Serif:

U+263A on Libertinus Serif and Liberation Serif

(This is also the case for the frowning counterpart, ☹ U+2639 "White Frowning Face".) 

Some links for context:
- https://unicode-table.com/en/263A/
- https://emojipedia.org/white-smiling-face/
- https://en.wikipedia.org/wiki/Emoticon#Unicode

waldyrious:
Fixes #183 by normalizing the shape of the following characters:

- ☹ (U+2639 WHITE FROWNING FACE) 
- ☺ (U+263A WHITE SMILING FACE)
- ☻ (U+263B BLACK SMILING FACE)

Most of the fonts used the same shapes for these characters, so I also reused the same corrected shapes. That said, in some cases adjustments had to be made:

- For the LibertinusKeyboard-Regular font, I scaled the shapes by 75%, placed them on top of the original ones, then rounded coordinates.
- For the italic fonts, I used the oblique transformation with value -12
- For the bold fonts, I wasn't able to parametrize the weight transformation to make the result match the original shapes (unlike the other two changes mentioned above, which produced a near-perfect match), so I manually edited the original shapes in the same manner I did for the regular characters.

(Note: This was my first time using FontForge. Fortunately my prior experience with Inkscape was somewhat helpful and I was able to learn enough of the basics to perform these edits. In any case, if I made any silly mistakes, that's probably why.)


khaledhosny:
Thanks!

I’m a bit concerned about the size, it seems to be too small compared to other symbols (and fonts), may be it should be at least enlarged to the size of U+262F?

waldyrious:
Yeah, I was wondering the same. I think it was made smaller due to the hair in the original shape, which I just removed without editing the rest of the shape. I will rework them to be larger, hopefully this weekend.

waldyrious:
Hey @khaledhosny — I've fixed the sizes to be compatible with U+262F ☯ as you suggested. I also made additional fixes:

- restored the fixed width of the characters in the monospaced font, which I had inadvertently messed with in my previous commit
- improved the shape of the smileys in the bold and bold italic fonts.
- added the keyboard button frame to the characters in the keyboard font

I also rebased the branch to incorporate the latest changes you made to master. In particular, the various updates you did to the keyboard font resulted in git conflicts with my changes, which I resolved manually. While doing so, I realized that one of the things you did was [remove](https://github.com/alif-type/libertinus/commit/7fb498a216d52ee44afd85afc8bbd18d7f3ada7c) the characters that didn't have the keyboard key frame around them, [including the smiley faces](https://github.com/alif-type/libertinus/commit/7fb498a216d52ee44afd85afc8bbd18d7f3ada7c#diff-35bfa3a8b31a01f3d9490d0e127f6a8dL24863-L25037). However, since I had already done the work of adding the frames to them, I kept them. I hope my work wasn't in vain!

I have a question, though. You may notice that in the updated patch, the otf fonts were not updated. That is because running `make` is now failing for me, which I assume is due to the recent changes to the build scripts. Can you help me debug what the issue may be? Here's the error I get:

```
(libertinus-env) ❯ make
      BUILD  LibertinusSans-Regular
       HINT  LibertinusSans-Regular
       SUBR  LibertinusSans-Regular
make: *** [build/LibertinusSans-Regular.subr.otf] Error 127
```

waldyrious:
Some screenshot to illustrate the changes:

| Before | After |
|--------|-------|
| screenshot-before | screenshot-after |

khaledhosny:
The build issue is probably because you are missing `afdko`, all Python modules required are listed in the `requirements.txt` file.

waldyrious:
> all Python modules required are listed in the requirements.txt file.

I did run `pip install -r requirements.txt` before running `make`. I'm not sure what could still be the issue, but maybe I'll open a separate issue for that.

PR #312, opened 2020-03-29
Update CONTRIBUTING.md

waldyrious:
- Add links for context and further information
- Use a custom virtual environment name (as discussed [here](https://github.com/alif-type/libertinus/commit/82236cb1565f4c58d55e078b4813fea5a7b86107#r38122587))
- Standardize references of file types
- Edit a sentence order to improve reading flow

@khaledhosny I'm wondering if we should also mention `xelatex` and `mutool` which `make` uses, even though they're not strictly required for the basic contribution flow. We could add a note at the end about optional dependencies, for example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /build
+libertinus-env
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,9 @@ Python dependencies is to use a [virtual environment][venv]:
     . libertinus-env/bin/activate
     pip install -r requirements.txt
 
+(There are also [optional dependencies](#optional-dependencies) which are not
+required for the basic contribution flow.)
+
 The source files are under the `sources` subdirectory. The `.sfd` files are
 FontForge source font format and should be edited with [FontForge][fontforge].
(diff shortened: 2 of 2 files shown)


waldyrious:
Also, we might want to suggest adding the environment name to the local [`.git/info/exclude`](https://help.github.com/en/github/using-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file) so that its content doesn't show as a bunch of uncommitted files.

Alternatively, we could recommend a specific environment name (e.g. "libertinus-env") and add it to the [tracked `.gitignore` file](https://github.com/alif-type/libertinus/blob/master/.gitignore).

khaledhosny:
> @khaledhosny I'm wondering if we should also mention `xelatex` and `mutool` which `make` uses, even though they're not strictly required for the basic contribution flow. We could add a note at the end about optional dependencies, for example.

Yes, please.



> Alternatively, we could recommend a specific environment name (e.g. "libertinus-env") and add it to the [tracked `.gitignore` file](https://github.com/alif-type/libertinus/blob/master/.gitignore).

Yes, that would be best. Please use `libertinus-env` and update the documentation and the `.gitignore` file.

PR #322, opened 2020-04-13
Don't make stuff, just work on the make stuff

alerque:
This is a rebase of #297 with all the Unicode coverage chart related commits removed, only the refactoring done to the Makefile itself is left.

If any of this needs explanation beyond what is in the commit messages for each change or any other reworking let me know.


Makefile
```diff
@@ -1,107 +1,121 @@
...
-SFD=$(FONTS:%=$(SOURCEDIR)/$(NAME)%.sfd)
-NRM=$(FONTS:%=$(BUILDDIR)/$(NAME)%.nrm)
-CHK=$(FONTS:%=$(BUILDDIR)/$(NAME)%.chk)
-OTF=$(FONTS:%=$(NAME)%.otf)
-SVG=preview.svg
-PDF=$(DOC)/Opentype-Features.pdf $(DOC)/Sample.pdf $(DOC)/Math-Sample.pdf
+NAME = Libertinus
+VERSION = 6.11
+
+MAKEFLAGS := -s -j$(shell nproc) -Otarget
```
waldyrious, line 4: I'd suggest adding a comment to this line explaining why these flags are added.

Makefile
```diff
@@ -1,107 +1,121 @@
...
+		$< 1> /dev/null || (cat $(BUILDDIR)/$(*F).log && false)
+
+preview.pdf: preview.svg
+	$(info         PDF  $@)
+	mutool draw -q -r 200 -o $< $@
+
+.PHONY: dist
+dist: check dist-clean $(OTF) $(PDF) $(SVG)
+	$(info         DIST  $(DIST).zip)
+	install -Dm644 $(OTF) -t $(DIST)
```
waldyrious, line 109: The previous code was more explicit and easier to understand. How about using the full command options, like `--target-directory` and `--mode` instead of `-t` and `-m`?

Makefile
```diff
@@ -1,107 +1,121 @@
...
-NRM=$(FONTS:%=$(BUILDDIR)/$(NAME)%.nrm)
-CHK=$(FONTS:%=$(BUILDDIR)/$(NAME)%.chk)
-OTF=$(FONTS:%=$(NAME)%.otf)
-SVG=preview.svg
-PDF=$(DOC)/Opentype-Features.pdf $(DOC)/Sample.pdf $(DOC)/Math-Sample.pdf
+NAME = Libertinus
+VERSION = 6.11
+
+MAKEFLAGS := -s -j$(shell nproc) -Otarget
+SHELL = bash
```
waldyrious, line 5: It's nice that you explain why this is done in the extended commit message, but a comment above this line would make the explanation more accessible.

Makefile
```diff
@@ -1,107 +1,121 @@
...
-		--input=$<                                                     \
-		--output=$@                                                    \
-		--version=$(VERSION)                                           \
-		$(if $(call nofea,$@),,--feature-file=$(GSUB))                 \
-		;
+$(BUILDDIR):
+	mkdir -p $@
+
+$(BUILDDIR)/%.otl.otf: $(SOURCEDIR)/%.sfd $(GSUB) $(BUILD) | $(BUILDDIR)
+	$(info       BUILD  $(*F))
```
waldyrious, line 66: The non-breaking space should be explicitly mentioned in a comment, otherwise it could easily be overlooked and inadvertently removed, or otherrwise cause confusion when searching or editing this file.

alerque: This would need to be commented all over the place, and the kind of people that would _understand_ such comments and bother with them are people that will probably catch this anyways. If not we can catch it on PRs. It isn't going to break people's builds if they muck around with this and get it wrong, it just won't line up the pretty messages. I think the cost of fixing in confusion and clutter this is higher than the risk of what could break.

Makefile
alerque: All most people are going to care about is the build time is `1 / CPU Count` of what it was before and the makefile is easier to read (less `@` clutter). Anybody that cares to understand MAKEFLAGS will get this, and no amount of commenting will help people that doesn't understand make. If you really want to know what they do check `man make`. I don't feel like duplicating that documentation is useful to future contributors to this project, only justifying the _change_ needed documentation.

Makefile
alerque: No. The `-t` flag is _very_ common in usage of `install` and far more common than its verbose brother. It's also the proper tool for the job, not `cp` in this case. I don't think explaining common shell syntax / usage is necessary in a context like this, that's what `man` pages are for.

This whole dist section is actually much _more_ explicit than it was before, and also more correct because the make targets have proper dependencies. I'm sorry make is a bear and kind of obtuse, but just because using it correctly makes it harder for you to follow doesn't make it worse. Phony targets that are just shell script placeholders may be "easier to understand" if what you understand is basic shell scripting, but that doesn't mean it's the best way to write a Makefile.

alerque:
> $ make
> make: touch: open: build: Is a directory

This error looks to me like you don't have commit 47c2048 — that's kind of what it is for. I just tested again in a perfectly clean clone `git -dxf` that has no build directory and it gets created properly when running `make` the first time.

alerque:
> My `make --version` is "GNU Make 3.81"

Oh. Sorry I'm seeing this last because I worked through your messages from the bottom up. I don't know about @khaledhosny but I have no intention of supporting that version of make. That was released in April 2006 (14 years ago) and the 4.x series of Make brought a lot of very good [changes](https://git.savannah.gnu.org/cgit/make.git/tree/NEWS). You'll need to get ahold of something a little newer than that.

Makefile
alerque: See [comment](https://github.com/alif-type/libertinus/pull/322#discussion_r407460766).

Makefile
waldyrious: I'm not sure if I came across as arguing for restoring the previous cp-based commands — I fully agree with using a proper tool like `install`, I was just suggesting using the flags that already come out of the box with it. Nothing about the new approach or its correctness would change; it just becomes more self-explanatory (which is helpful because, as you say yourself, make is not exactly the most intuitive to work with).

IIUC short options are most helpful when typing commands by hand, but in a script that's already written out, there's no downside in using the full option name, don't you agree?

waldyrious: (If that's unacceptable to you, then at least consider splitting `-D` and `-m`, for readability, and adding a space after `-m` like you did with `-t`: `-D -m 644`.)

waldyrious:
> I have no intention of supporting that version of make

Ok, that's helpful to know. What would be the minimum version you'd say one would need? I'd like to add this to CONTRIBUTING.md as is currently done for Python and mupdf-tools.

Makefile
alerque: First, `install` is not part of GNU Make, it is part of GNU's core utilities, which was in turn made to imitate the older BSD command of the same name. Tho long option names are a GNU only thing, they are not implemented in the BSD versions of make.

Secondly, even if the long options were universal (and in our case we are kind of assuming GNU tools anyway), there are several downsides to long arguments in scripts. One of them is that they tend to be so verbose they make simple lines that would otherwise be short enough to read so long that they wrap in the editor. This makes them _harder_ to read, not easier. Secondly the short options are (in this particular case) much more commonly used in the wild. You will see them in thousands of packaging scripts using these short options, and they are _more_ recognizable to anybody that works with these things than the long options. And less specific to this case, short options are also preferred in scripts where the CLI usage is part of many people's muscle memory. There is no sense in writing `tar xfz ...` on the CLI and `tar --extract --ungzip --file=...` in a script. Heck had to use `man` just to figure out what those are, and I would have to do the same for `install`, whereas thousands of times seeing and writing the short options have them drilled in my head and fingers.

alerque:
I'm guessing 4.1 might be the actual minimum, but 4.2 has been around for 4 years and it is the oldest version I'm testing regularly with.

waldyrious:
Cool — I'll see if I can test with 4.1; otherwise, it might be safer to just require 4.2.

alerque:
@waldyrious Partly out of curiosity, partly out of incredulity I have to ask: how do you even end up with a system with a 14 year old version of GNU make? I know at some point more recently than that OSX was shipping with BSD make, and that sometime in the last couple years they started shipping GNU make. Does this mean that before than they shipped GNU make, then switched to BSD and now back again? That's _such_ an old version I can't figure out how that would end up in a modern Mac OS system. Is your OS itself up to date? Did you install GNU tools through some other channel?

Makefile
alerque: > then at least consider splitting -D and -m

Popular usage has these grouped together as `-Dm666`. I grepped the projects directory on my system (which has package repository sources files for several Linux distributions) and found several hundred thousand usages for `install -DmNNN` and just a few hundred with them split as in `install -D -m NNN`.

_YOU_ may not be familiar with the usage, but anybody with experience with `install` from other contexts is far more likely to have run into the combined usage, just like is is far mor common to see `tar xfz` than `tar -x -f -z` even though both are valid.

Makefile
khaledhosny: I prefer to set parallelism manually, I hate tools that always build in parallel without asking.

Makefile
khaledhosny: I’m fine with using no break space, my editor is always configured to highlight these (as I often use them elsewhere).

Makefile
alerque: Really? When is it better to _not_ build in parallel? I'm constantly annoyed at having to manually add these flags and patch builds so they can be done in parallel. Few modern systems only have single core CPUs and except in cases where there are dependency race conditions (not the case here) there is no real benefit to single threading the entire process. For this project on my old 3 core AMD system it is almost exactly a 3× speedup building these fonts. Why wouldn't you want that?

khaledhosny: When I’m developing I often make broken changes, failing at the first font is better than having to wait for all parallel jobs to fail (the broken recipe might not be the first one, so build will not fail fast)..

Makefile
khaledhosny: Grouped short options are fine and very common, so I have no objection here.

Makefile
alerque: Given the way make dependency tries work, `make` should still fail faster with more jobs running than less, and it will not run any new jobs after one has failed. The output grouping option I setup should also make sure the failed job's output is last. I'll mockup a proof of concept to check that now.

alerque: Using GNU-Make, the make process will throw the error and output messages from the failed recipe immediately on failure. It does let any other unfinished jobs that were started also run and output (clearly marked), but it will not run any new jobs. If what you want is fast failure to see what you broke, defaulting to parallel builds this way is to your advantage. It will still get you the least time from invocation to error condition whether the error is in the first job spawned or the 50th.

khaledhosny: Say I made something stupid that breaks autohinting, compiling the font and the features (the first recipe) will not fail but autohinting will fail. With one job only the firat font will fail and I’ll see the failure, with multiple jobs all the fonts will be built then fail one by one during autohinting, which is much much slower.

khaledhosny: (`nproc` returns 6 for me, so that is 5 more jobs to wait for).

alerque: Hm, that is actually an interesting scenario. I see what you are talking about now. What was faster in the end would depend on what the fix was, if going back to redo the first build step was required for all fonts, it could be slower over all. If the fix only affects the later stage _OR_ only effects some fonts and not others, this will still be faster. In fact unless the source files for all the faces are being changed at once, parallel should be better. Remember it doesn't always start at the beginning, it will restart with whatever the last incomplete thing is.

Would your use case help to have a separate terminal open with `git ls-files | entr -c make`? and add `-j1` only when you are doing something that modifies all the source files in parallel? Anything editing only one source file at a time will still be faster with parallel builds.

khaledhosny: That is too clever for me. I usually build with `make -j` and unless I‘m doing some change to build scripts, and before committing I always build with make -Bj` because I’m paranoid and want to make sure I have a clean rebuild.

waldyrious:
I'm using a relatively recent macbook, and the latest macOS version (Catalina).

```
❯ /usr/bin/make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

Just to be sure this wasn't installed by some third-party tool — I get the same output with `/Library/Developer/CommandLineTools/usr/bin/make --version`.

The Homebrew-provided make (under `gmake`) shows the following:
```
❯ gmake --version
GNU Make 4.3
Built for x86_64-apple-darwin19.2.0
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

alerque:
> The Homebrew-provided make (under `gmake`) shows the following:

I have no idea what Apple is thinking. It was bad enough they vere shipping BSD make (which is a different beast with even more issues). That they switched to shipping GNU make but ship a 14 year old version in their "modern" OS is just bonkers. I'm assuming that means they probably use it internally to build parts of OSX and haven't updated their internal build process in 15+ years. So that's spiffy.


I'll hazard a guess you don't have any build errors if you use `gmake` from Homebrew.

khaledhosny:
Apple stopped updating GNU software when it switched to GPLv3.

Makefile
alerque: Personally I am completely unconvinced. I think the majority of users will have no idea they can/should set this manually on every run and that the majority of use cases would benefit from it. It drives me crazy having to add it all the time. For the rare cases during development cycles when a different option would provide a faster feedback loop, you know what arguments to pass (and I would actually do it differently by leaving the job server untouched and setting `make FONTS="..."` instead to manually limit the font set I wanted to work on during testing).

At this point if I've failed to convince you and this is blocking merging this PR I will remove it, you are after all the primary developer. Just let me know.

waldyrious:
> I'll hazard a guess you don't have any build errors if you use gmake from Homebrew.

I actually do, but it seems to be due to it not finding `sfntedit`

    bash: line 2: sfntedit: command not found

Not sure why this didn't come up before. In any case, it seems to be another entry that should be added to the optional dependencies in CONTRIBUTING.md, right @khaledhosny?

alerque:
> I actually do, but it seems to be due to it not finding `sfntedit`

The `sfntedit` binary should be provided by [AFDKO](https://github.com/adobe-type-tools/afdko). Yes that should be mentioned.

waldyrious:
> The `sfntedit` binary should be provided by AFDKO. Yes that should be mentioned.

Ah, that is already in the `requirements.txt` (#310), which reminded me that I wasn't actually using the virtual environment. After activating it, `gmake` works as expected without any issues. Sorry for the false alarm.

Makefile
khaledhosny: Yes, please. Parallel make by default breaks my workflow.

Makefile
```diff
@@ -1,107 +1,121 @@
...
+TOOLSDIR = tools
+
+PY ?= python3
+BUILD = $(TOOLSDIR)/build.py
+NORMALIZE = $(TOOLSDIR)/sfdnormalize.py
+
+# Default to explicitly enumerating fonts to build;
+# use `make ALLFONTS=true ...` to build all *.sfd files in the source tree or
+# use `make FONTS="Face-Style" ...` to make targets for only particular font(s).
+ALLFONTS ?= false
```
khaledhosny, line 22: Why do we need this at all? It is not likely that we will be adding any new styles any time soon, and even we did not updating the list here will be obvious mistake as the fonts wont be built.

alerque: For your workflow, you probably don't need this. It serves two scenarios. First people like me that are messing around with the font can easily copy one of the source files to a new name and muck around in it while comparing it to the original, and we can have an option to build everything from sources without having to explicitly list what those are. Second it makes for a bit more flexible overall setup if copy/pasting and adapting for a new font project that might not be as set in stone yet as this one is. Your preferred workflow of building exactly the fonts specified for this project by default is retained.

khaledhosny:
GitHub is saying there is a merge conflict and I can’t merge the PR.

alerque:
Have you had the tab open a long time? If so hard refresh. One way or another it lies. there has been no activity on the master branch since I last merged _from_ master into this branch, it's impossible that there would be a conflict merging back. Also I just tested it on the command line and ... as expected ... no issues.

khaledhosny:
Nope, I’m opening it anew every time. Won‘t let me to merge and I’m sure it is not a cache issue.

khaledhosny:
OK, only rebase and merge was disabled, merging with a merge commit (which happens to be what I actually wanted here) was fine.

alerque:
@khaledhosny Ya I wouldn't expect it to be able to rebase this, I manually resolved merge conflicts against the stuff coming from master since I started and rebase wouldn't know what to do with those. Glad it worked in the end.

PR #330, opened 2020-07-05
Adopt, update project meta data

alerque:
Having adopted this project as the current maintainer, this updates the license, readme, etc. with current meta data.

@khaledhosny I see you removed your information from the Github FUNDING‌ meta data file. With your permission I would actually like to restore that. Even though you have passed on the maintainer role and I don't think anybody has a right to expect anything from you as far as future contributions you have done the lion's share of the work over the last 8 years making the font what it is. Anything I do will only be building on that legacy. Even in the event somebody comes, likes what they see now, and wants to sponsor the project I think it's only fair they consider tipping you for your past work. It may or may not ever amount to anything financially, but not pointing people in that direction doesn't seem right to me ... unless of course you are actively uncomfortable with it. In that case I will of course respect your wishes. I just wanted to voice that my wish would be for your work to be recognized and and potential supporters pointed your direction as much or more than my own.

To that end would you mind voicing your approval or disapproval of my revert of your commit (e37134b) and extra mention in the updated README (f94417f).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,24 +1,19 @@
-Contributing
-------------
+# Contributing
 
-To build the fonts, you need [GNU Make][make] and [Python 3.6+][python], as well
-as [a few Python packages](requirements.txt). The cleanest way to install the
-Python dependencies is to use a [virtual environment][venv]:
+To build the fonts, you need [GNU Make][make] and [Python 3.6+][python], as well as [a few Python packages](requirements.txt).
+The cleanest way to install the Python dependencies is to use a [virtual environment][venv]:
 
     python3 -m venv libertinus-env
     . libertinus-env/bin/activate
     pip install -r requirements.txt
 
-(There are also [optional dependencies](#optional-dependencies) which are not
-required for the basic contribution flow.)
+(There are also [optional dependencies](#optional-dependencies) which are not required for the basic contribution flow.)
 
-The source files are under the `sources` subdirectory. The `.sfd` files are
-FontForge source font format and should be edited with [FontForge][fontforge].
-The `.fea` files are [OpenType feature files][fea] and should be edited by a
-plain text editor.
+The source files are under the `sources` subdirectory.
+The `.sfd` files are FontForge source font format and should be edited with [FontForge][fontforge].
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -28,45 +23,37 @@ We keep the generated fonts under version control, so the last step is to run:
...
+The `make` command will emit errors if these are not present, but as long as the font generation stages finish correctly, these errors can be ignored.
+However, if you do wish to update the preview files, these tools must also be installed.
 
-#### Linux
+### Linux
 
-On most Linux distros, the `xelatex` binary is provided by the
-[`texlive-xetex` package][repology-xetex], and the `mutool` binary by
+On most Linux distros, the `xelatex` binary is provided by the [`texlive-xetex` package][repology-xetex], and the `mutool` binary by
 [`mupdf-tools` or `mupdf`][repology-mupdf].
```
waldyrious, line 39: I suspect you may have missed this line break, @alerque.

alerque: Good catch.

khaledhosny:
> @khaledhosny I see you removed your information from the Github FUNDING‌ meta data file.

I find it rather misleading to ask people for donations while I’m not involved in the project any more, so I’d prefer to keep it this way, thank you.

alerque:
@khaledhosny That's for the reply. Even if I disagree about it being misleading (I feel it's misleading to _not_ point the credit more towards you for past work than me for current nothing) I'll respect your wishes in this. If you ever change your mind feel free to comment again or open a PR to add current links back.

OFL.txt
```diff
@@ -1,5 +1,4 @@
-Copyright © 2012-2017 The Libertinus Project Authors, with Reserved Font Name
-"Linux Libertine", "Biolinum", "STIX Fonts".
+Copyright © 2012-2020 The Libertinus Project Authors
```
khaledhosny, line 1: I believe that the reversed font name clause should be kept even of this fork is no longer using this name, since anyone forking it can’t use any of these reserved name.
https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL-FAQ_web&_sc=1#726d6952

alerque: Thanks for this correction, I went to the FAQ looking for this and apparently missed it.

PR #331, opened 2020-07-06
Drop generated binary fonts from repository

alerque:
With all due respect to previous maintainers (and indeed the vast majority of designers dumping their fonts into Git) this is and always was an incorrect use of version control. The problems with it are legion: massively inflated repository size, compete inability to use VCS tooling for things like diff, merge, rebase, revert, etc. The two main "problems" this work flow solves can easily be solved with the proper tools for the job: reproducible build scripts and archives of the generated fonts.

This font project in particular is in an excellent position to take this leap because the builds are already reproducible and the primary venue for releases (Github) allows attaching archives to releases.

----

Handling past releases is a trickier problem. I'll open an issue to discuss that.


waldyrious:
@alerque I agree that the binary files do not need to be tracked if they're included as archives in the releases; but I would suggest to take this opportunity to let the root directory become less cluttered as the .otf files are removed, by not [moving the .sfd files to it](https://github.com/alerque/libertinus/pull/331/commits/4c27b5ba2fb39d8b079c5c1b5d019f6b4564760a). The list of files at the root level is already long enough :)

Compare how the repo would look with this PR (left) vs. keeping the `sources` directory (right):

alerque:
Thanks for the feedback @waldyrious. I was actually planning on trimming down some of the other file noise at the root level as well (no reason to have non-essential documentation in the root _and_ a documentation folder, the log files can be combined, etc.) but I'm not convinced on having a "source" directory. I know there is precedent both ways in the programming world. Rust for example pretty much railroads you into using a `src` directory. Python strongly encourages it. Other language ecosystems make it difficult to do, others flat out require sources at the root.

I would actually be interested in hearing you take this up as as issue on the [fontship](https://github.com/theleagueof/fontship). I have an RFC there for distribution package layouts, there should probably be one for source repo layouts as well. I plan to make it flexible enough to configure, but defaulting to and encouraging a simpler layout than what most font projects today use (sources & fonts each in directories). For projects with Glyphs sources being Canonical the typically only have one or two source files anyway. Having this in a subdirectory is almost silly. For UFO and SFD there tend to be a few more sources, but I think Libertinus is pushing the upper limit. I'm not particularly worried about the Github display as you can deep link to the readme anyway.

Would you mind making a case for your preference over there so I can get it defaulting to something sane?

fitojb:
Google Fonts requires some structuring: https://github.com/googlefonts/gf-docs/tree/master/MasteringNewFamiliesForGF#repo-layout

alerque:
@fitojb Google Fonts and I do not see eye to eye here. While they have also been a great benefactor to web typography and many free font projects I also blame them for perpetuating some of the terrible font project management practices that are out there now. I am aware of their expectations because of their terrible tooling (have you ever tried to _use_ `gftools`? Every subcommand has different expectations about output locations!) and don't consider that a valid argument to maintain the dumb status quo. I'm not blaming anybody specific at Google but the overall pattern is dumb and I am on a campaign to change it. Writing a tool that works better and makes life easier for projects collaborating on font design such as open source projects is the step I of that. Campaigning to get Google Fonts to stop encouraging the terrible practice like commiting binaries to repos (which they don't use anyway, they run the build toolchain themselves) is step II.

khaledhosny:
I originally did this because font building tools are unreliable (mostly FontForge), and I needed away to catch regressions in these tools. Things should be better now, but I still catch tool bugs and unintended changes this way (especially when returning to a project after a relatively long period of in activity, there are almost always changes the the generated fonts that I need to carefully evaluate).

Pinning the build dependencies vigorously (not what I’m doing now, where I pin only the packages I explicitly use but not their dependencies) might achieve the same effect as one can then reliable reproduce the last build, which was not easy before.

I’m not strongly opposed to this change, but I have seen how the sausage is made and I have no trust in the process.

The other reason, is that I often don’t make releases for a long time (for various reasons) and people would use the files from the repo in the mean time.

alerque:
Thanks for that input @khaledhosny. I do get the benefit of comparing the generated files with past versions. To that end I'm working on making `fdiff` play nice with Git and plan on facilitating this in `fontship`, but I think the best way to accomplish the same result is to use the previously generated archives and have the tooling help make that process easier.

I do something similar for my book publishing work: book repositories have the SOURCE of the book (Markdown) but not the generated PDF's for each commit during the authoring, editing, and typesetting process. I can still compare PDF's for all previously build book versions using tooling that fetches them from a release archive and uses visual PDF diff tools to show differences. Having the PDF's for every commit to the source would be _HUGE_, but it isn't hard to write tooling that manipulates archives wherever they are stored without getting in the way of other more naturally suited VCS functions.

moyogo:
Have you considered [git LFS](https://git-lfs.github.com/)?

alerque:
@moyogo Not in a while. The last time I used LFS was years ago. Off the top of my head though I can't think what advantage it would bring to the table over just not tracking bulid artifacts in repo at all. I'd be grad to hear an elaboration, but maybe [this issue](https://github.com/theleagueof/fontship/issues/47) would be a better home for the discussion. Whatever way I agree to take this repository will be the same way I set the defaults in `fontship`.

alerque:
Hey all ... I'm not yet convinced about the merit of keeping the `sources/*` layout, but I also don't see any reason to hold up this PR while that is discussed. I would appreciate hearing input on that one over on fontship, but I've backed the relevant commit out of this PR for now. I can still put it through later and this way won't cause file rename churn in the history (another sore point for most Git tooling).

Issue #332, opened 2020-07-06
How should we handle the binary artifacts in Git history?


alerque:
As the new maintainer I am making the executive decision to stop adding generated binary artifacts to the Git repository (see #331). To me the benefits of being able to use tooling designed for the purpose of managing sources far far outweigh the expediency of having matching artifacts¹ & sources side by side at the expense of not being able to efficiently use tooling).

What to do with the repository history to date is a much trickier problem. As of 6.12 the Git object store for the repository is 333 MB. That's hundreds of megabytes of baggage you are forced to download and shuffle around to do anything meaningful with the source (sure you _can_ get a sparse checkout with modern Git versions, but who knows how to do that!).

1. One option would be –given that this won't keep growing in the same way now— to cut our losses, say the 333 MB of heft is just our baseline and suck it up and keep it.

2. Another option would be to keep a copy of the current main branch and tags around on Github untouched as a legacy source, but add a new branch that is rebased using filter-branch to remove all the artifacts. Future releases could be tagged against this branch, and it could be set as the default branch. This would mean most people cloning the repo would only get the lightweight development branch while people pulling `--all` branches would git a surprise payload.

3. The final main option I see is to do the same rebase as in option 2, but drop the current branch entirely. This would mean re-tagging all the releases.

I really don't like option 3, there is too much history: links across the Internet, checksums for existing releases, cross-linked issues, links to commit SHA's, etc.

Option 1 is the least friction, but I'm not too happy with permanently dooming the repository to dragging around the artifacts.

I propose experimenting with option 2 and at least seeing what it would look like. How big would the git object store be for a branch with all the artifacts filtered out? How far back are the releases fully reproducible? Can we script something to package all the existing Git artifacts and attach them to their respective tags? Would there really be any data loss this? It looks like at least all the 6.x releases since 2016 have the artifacts attached to the releases already, so this is only relevant to older tags.

¹ <sub>Going forward I'll be porting the SFD build process to use [`fontship`](https://github.com/theleagueof/fontship). One side effect of this is that automatically generated build artifacts will be available for every commit to Github (not just releases) so going forward the equivalent of being able to get font artifacts at arbitrary non-release commits will still be possible anyway.</sub>

fitojb:
I’d be happy going with option 3, but number 2 sounds like a good compromise. People are going to have to update links to this project due to the change of maintainer anyway (don’t trust GitHub to maintain any redirect forever!), so it seems like it’s the right time to make the repo leaner (y)

[...]

PR #334, opened 2020-07-07
Build with Fontship

alerque:
This is a work in progress that uses as development branch of `fontship`, I'll wait to merge it until:

- [x] All font artifacts previously in repository are generated for every commit
- [x] Font artifacts are diffed with previous version to make sure nothing changed
- [x] Upstream `fontship` has a tagged release that supports SFD based projects,
- [x] Rules (at least temporary ones) are setup for regenerating specimens,
- [x] SFD normalization is supported again,
- [x] The check target works again.

...but it's to the point now where it's generating fonts (which can be done locally or downloaded from any pushed commit's CI job artifacts).


Makefile
```diff
@@ -41,3 +41,11 @@ $$(BUILDDIR)/$1-%-instance.otf: $$(BUILDDIR)/$1-%-static.otf
 	$$(PYTHON) -m cffsubr -o $$@ $$<
 
 endef
+
+$(DOCSDIR)/preview.pdf: $(DOCSDIR)/preview.tex $(STATICOTFS) | $(BUILDDIR)
+	xelatex --interaction=batchmode -output-directory=$(BUILDDIR) $<
```
khaledhosny, line 44: I use XeTeX for such a sample thing because of the math equations, and sadly there are no other command line tools that can render OpenType math other than *TeX.

alerque: Yup, hence why I re-instated this rule after having removed under the assumption I could do something else. Most of the example actually renders in SILE's experimental math support branch, but it's not merged yet or anywhere close to ready. I'm tabling this for now until a better solution exists.

alerque:
Merging this now, but there are a couple more steps before tagging a release.

As a heads up to anybody watching, when I do tag the next release it will be `7.000` (with three minor version digits according to [OpenFV versioning format](https://github.com/openfv/openfv)). The font should a 100% metrically compatible drop in replacement with the 6.12 release with not even a single glyph or kern change, only everything outside of the actual font has changed: repository structure (#332), build system (this PR), and output zip layout all reflect "new management".

The release ZIP layout will now, based on [current Fontship defaults (that are still RFC)](https://github.com/theleagueof/fontship/issues/19), the contents of the distribution will have shifted slightly, rather than being at the top level in the ZIP the files everybody is going to be looking for are under `static/OTF/*.otf`.

Besides the version number format change and release package layout change, font packagers shouldn't have any trouble updating. After that we can start chipping away at actual font issues.

PR #344, opened 2020-08-17
Update font foundry information since adoption

alerque:
Planning to merge after #334 as a test of the build system.

### DIFF
--- a/sources/LibertinusKeyboard-Regular.sfd
+++ b/sources/LibertinusKeyboard-Regular.sfd
@@ -42,7 +42,7 @@ HheadDOffset: 0
 OS2CapHeight: 754
 OS2XHeight: 754
 OS2FamilyClass: 2063
-OS2Vendor: 'ALIF'
+OS2Vendor: 'QUE'
 Lookup: 4 0 1 "'liga' Standardligaturen" { "'liga' Standardligaturen 1"  } ['liga' ('DFLT' <'dflt' > 'cyrl' <'dflt' > 'grek' <'dflt' > 'latn' <'AZE ' 'CRT ' 'TRK ' 'dflt' > ) ]
 MarkAttachClasses: 1
 DEI: 91125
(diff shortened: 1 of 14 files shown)


sources/LibertinusKeyboard-Regular.sfd
```diff
@@ -42,7 +42,7 @@ HheadDOffset: 0
 OS2CapHeight: 754
 OS2XHeight: 754
 OS2FamilyClass: 2063
-OS2Vendor: 'ALIF'
+OS2Vendor: 'Alerque'
```
khaledhosny, line 45: OS/2 vendor is a 4-letter tag, and are customary [registered with MS](https://docs.microsoft.com/en-us/typography/vendors/) (though it is not a requirement, but some tools use the registry and the process is rather simple).

alerque: Registered with M$FT? Of course it is. *&lt;mumble splutter grumble&gt;*.

> The Vendor ID identifies the company responsible for the marketing and distribution of a typeface. It is reasonable to assume that there will be 6 vendors of ITC Zapf Dingbats for use on desktop platforms in the near future (if not already). It is also likely that the vendors will have other inherent benefits in their fonts (more kern pairs, unregularized data, hand hinted, etc.).

Leave it to have Microsoft list "unregularized data" listed as a *benefit*! In any case since they are so font of irregularity I've attempted to register a three letter VendorID for my own personal use.

Thanks for the tip.

alerque: Apparently 3 characters gets a pass, they approved me surprisingly quickly:

https://docs.microsoft.com/en-us/typography/vendors/#q

khaledhosny: Yes, they respond quickly and don‘t ask question. It is still a 4-letter tag, but it get padded with spaces.

PR #353, opened 2020-08-21
space(latin): Fix side bearings for +dlig 'tz'

### LINKED ISSUE #342: rare issue with Libertinus Sans Bold: r and old-style ligature tz are overlapping in 'rtz'
With LuaLatex or LibreOffice with discretionary (old-fashioned) ligatures enabled you get this:

image

Obviously, the space between r and tz must be larger.

Best,
Robert

alerque:
Closes #342.

How does this look @rriemann?

I changed the side bearings for both Serif Display and Sans Bold.

image


rriemann:
Looks good. Though in Sans Bold, the difference in Wiertz between *r* and *t* and on the other side left and right side of *e* looks disproportionate (latter too large). Though I guess this is off-topic in this pull request.

alerque:
What I did for each face I adjusted is copy the left side-bearing for that face's ‘t’ and applied it to the ‘t_z’ ligature. Since the left edge of the glyphs are pretty much the same I figured this would resolve this, but I to see in the Sans Bold variant the `rt` kerning looks too tight compared even to the non-ligature `rt`.

alerque:
On closer inspection I think what you're seeing @rriemann in the preview above is actually a consequence of hinting an rasterization at the size I took the screenshot. A closer inspection suggests they are much the same as I intended:

image

PR #356, opened 2020-08-27
feat(symbols): Add 𝄆 and 𝄇 (U+1D106–7)

### LINKED ISSUE #339: Add support for U+1D106 and U+1D107 (musical symbol left|right repeat sign)
I'm [developing a tool](/gyuris/openlyrics-prototype) for for Christian worship leaders to make chord sheets from XML using CSS. Currently I use previous version of Libertinus (Linux Libertine/Biolinum). I have to format repeating lines with musical repeat signs:
```
𝄆 Repeating line 𝄇
```
Would it be possible to support these two glyphs in Libertinus?

(I'm not a font designer, just a user. I made `Opentype-Features.fodt`)

alerque:
Closes #339

Hey @gyuris can you give this a scan? I only did the Serif Regular face for now. If this looks sane I'll do similar alignments for the other faces before merging this.

image

The image is for convenience, but you should be able to download actual fonts (from any PR the "Checks" tab, click on either Fontship run and download the artifacts). Here is the [artifacts ZIP](https://github.com/alerque/libertinus/suites/1108692350/artifacts/15556931) for this PR as of now.

The dots are composed of a colon `:`, the thin bar is the weight and position of a pipe `|`, and the thick bar is just a doubling of that. The space between the bars is from the double bar `∥`.


gyuris:
I'm impressed! It was much faster than I had hoped. I'm absolutely satisfied. Thanks!

Sample:

new_symbols

alerque:
Great I'm glad that will work for you. Expect to find these across most of the faces in the next release.

I'll likely be using them as well to replace my usage of `//` in lyics to indicate repeats, as show here (excerpt from bulletin typeset with LaTeX & Lilypond using Libertinus fonts throughout—and yes we have bulletins with music and bulletins with just lyrics).

image

Note my context will actually be italics. I had assume these glyphs should remain upright, but there doesn't seem to be a clear precedent for that. Symbola, Bravura, and Noto Music don't have italic styles at all. FreeSerif does, but these glyphs are entirely absent in it, only the Roman face has them. Vollkorn on the other hand has them slanted!

image

Honestly it looks to me like that will be the best bet. People that don't want them slanted can just use Roman right?

gyuris:
I agree with you: slanted glyphs are good for italic text.

alerque:
Here is Libertinus Serif Roman, Italic, and Bold. Rebuilt fonts should show up on checks as artifacs in a few minutes if you want to use them. I'm going to bed ;-)

image

alerque:
Having redone the weight of the period (and hence the colon) as well as the vertical alignment of the colon (see #367) I've merged those changes into this branch so that the derivative glyphs added here will have the matching weights and positions.

I've also expanded the character to cover all the relevant styles (all except the Keyboard, Initials, and Math faces).

Here is a sample of where this is at now:

image

<details>
<summary>(SILE source)</summary>

```sile
\begin{document}
\neverindent
\nofolios
\language[main=und]
\script[src=packages/grid]
\grid[spacing=50pt]
% \grid:debug
\begin{raggedright}
\begin{script}

local files = {
	"LibertinusSerif-Bold.otf",
	"LibertinusSerif-Italic.otf",
	"LibertinusSans-Regular.otf",
	"LibertinusSans-Bold.otf",
	"LibertinusSans-Italic.otf",
	"LibertinusMono-Regular.otf",
}

		func(options, i, font)
	end
	SILE.call("par")
end

loop(function(options, i, font)
	SILE.call("font", options, { "𝄆 "..font.." 𝄇", "   ", "𝄆|𝄇", "   ", "𝄆m𝄇m𝄆m𝄇" } )
	SILE.call("break")
end, { size = "16pt" })

\end{script}
\end{raggedright}
\end{document}
```

</details>

alerque:
Hey @gyuris would you be in a position to download the latest font artifacts built from this PR and give them a quick test drive before I merge this. There shouldn't be much change from what you had before. I did fix the side-bearings just a touch, how the glyphs are built has been redone, and I've expanded support to cover the whole family. This should be ready to merge now if all looks good.

gyuris:
Thanks @alerque. It works well. It is much more than I had hoped for. You are awesome.

PR #367, opened 2020-10-03
add(modifier): Add ‘꞉’ [U+A789] MODIFIER LETTER COLON

### LINKED ISSUE #364: Link Uni A789, Modifier letter colon to Uni 003A, Colon
There's a Unicode slot for ꞉ (Unicode A789), which is exactly the same glyph as the regular colon. It's purpose is purely technical, particularly in linguistic texts. It is thus sufficient for A789 to resemble 003A, the usual colon.

I'd be happy to create a PR for this fix, but I'm unsure how to link to Unicode characters in FontForge (or any other software), so that future changes to 003A would take effect on A789 as well.

alerque:
Closes #364.

This *was* going to be my attempt to add `U+A789` to the relevant fonts (all except Math, Keyboard, and Initials). However given [this comment](https://github.com/alerque/libertinus/issues/364#issuecomment-703153120) it sounds like this is the wrong way to do this and I'm going to have to actually do some drawing.


alerque:
Okay I've redrawn these using transformed references to period instead of composing them of colons. @kopeckyf if you could give this a test drive before I merge that would be great. You can find font files in this PR under the 'Checks' tab → Fontship → artifacts.

Extra thanks to @moyogo for the link to the actual Unicode proposals and examples. Hopefully this is looking more suitable than a bunch of colons now.

image

<details>
<summary>(SILE code)</summary>

```sile
\begin{document}
\neverindent
\nofolios
\language[main=und]
\font[size=25pt]
\begin{script}
SILE.registerCommand("sample", function(options, content)
	SILE.call("font", { filename = "/home/caleb/projects/libertinus/"..content[1] }, {":꞉ ", content[1]} )
	SILE.call("hfill")
	SILE.call("break")
end)
\end{script}
\sample{LibertinusSerif-Regular.otf}
\sample{LibertinusSerif-Semibold.otf}
\sample{LibertinusSerif-Bold.otf}
\sample{LibertinusSerif-Italic.otf}
\sample{LibertinusSerif-SemiboldItalic.otf}
\sample{LibertinusSerif-BoldItalic.otf}
\sample{LibertinusSerifDisplay-Regular.otf}
\sample{LibertinusMono-Regular.otf}
\sample{LibertinusSans-Regular.otf}
\sample{LibertinusSans-Bold.otf}
\sample{LibertinusSans-Italic.otf}
\end{document}
```

</details>

kopeckyf:
Thanks @alerque, much appreciated. I noticed that there isn't that much distinction between the regular and semibold variants. However, they also do not differ very much for the usual colon, so I don't think that's too much of a deal. The other thing I saw was that the kerning following the italic variants was just a little bit too wide. Do you think it would be a good idea to reduce it just a bit so that the modifier letter colon is placed exactly between two letters?

alerque:
Thanks for the feedback @kopeckyf.

Yes I noticed the weight problem in Semibold too. I'm going to fix that. I'm also going to fix the vertical alignment of both the color and this modifier. Currently the dot placement is not consistent at all. Mostly they are aligned on the baseline and then the weights grow from there, but even that is not true for all faces and styles. I'm going to completely redo that so that the center point of dots is at a consistent height across the weights. This will bring the regular colon up off the baseline just a touch in all weights except bold. I'm trying to do this without changing horizontal metrics at all, although the italic styles are a bit tricky in that regard. At least one of the italic styles wasn't really matching the italic axis either (most were close, one was much less sloped than the axis).

I'll look into the horizontal metrics of the modifier letter colon. I probably used the side bearings from the colon which are not balanced, but since the modified is a letter part of words those offset bearings are probably not appropriate.

alerque:
Okay @kopeckyf here is round two. I think all the things discussed so far are covered. I would appreciate it if you can give it another test drive. New artifacts should arrive in a few seconds under this PR's 'Checks' tab, then the Fontship builds.

Here's a sample of the colon before changes (v7.000), plus the colon and U+A789 situation after changes:

image

<details>
<summary>(SILE code)</summary>

```
\begin{document}
\neverindent
\nofolios
\language[main=und]
\script[src=packages/grid]
\grid[spacing=25pt]
\grid:debug
\begin{raggedright}
\begin{script}

local files = {
	"LibertinusSerif-Bold.otf",
	"LibertinusSerif-Italic.otf",
	"LibertinusSans-Regular.otf",
	"LibertinusSans-Italic.otf",
	"LibertinusMono-Regular.otf",
}

local both = true

	options = options or { size = "125pt" }
	if both then
		SILE.call("font", { size = "10pt", family = "Libertinus Sans" }, { "Colon before: (Display, Serif, Serif Italic, Sans/Italic, Mono)" })
		SILE.call("break")
			options.filename = "/home/caleb/projects/libertinus/Libertinus-7.000/static/OTF/"..font
			func(options, i, font)
		end
		SILE.call("break")
		SILE.call("font", { size = "10pt", family = "Libertinus Sans" }, { "Colon after: (Display, Serif, Serif Italic, Sans/Italic, Mono)" })
		SILE.call("break")
	end
		func(options, i, font)
	end
	SILE.call("par")
end

loop(function(options, i)
	local str = ":"
	if i == 1 or i == 4 or i == 7 then str = str .. " " end
end)

SILE.call("bigskip")
both = false

loop(function(options, i)
    local str = "꞉"
	if i == 1 or i == 4 or i == 7 then str = str .. " " end
end)

loop(function(options, i, font)
	options.size = "20pt"
	SILE.call("font", options, {"o꞉o꞉o :꞉ "..font} )
	SILE.call("break")
end)

\end{script}
\end{raggedright}
\end{document}

```

</details>

alerque:
Expect this in the next release. Of course until then feel free to use development builds. The latest Git `master` branch always has artifacts attached to the checks. Let me know if any issues crop up.

PR #368, opened 2020-10-03
Normalize FontForge sources

alerque:
“But they *were* normalized!” I hear you say.

I thought so too.

There are two separate changes here, both of which I have questions about that should be answered before I merge this.

1. When I inherited this project in July, I could open the font files in FontForge, save them, run the normalizer, and everything would be clean. Now that is not the case, and I don't understand why. I'm still using FontForge 20200314, so that hasn't changed. My distro did recompile it on August 30th (against new GTK libraries I think) but I can't find any reason this should be happening. Commit fca3f1e is the result of opening every SFD file, saving it, closing it, and re-running the normalizer. @ctrlcctrlv maybe you have an idea what is causing that particular set of changes? Is this something I should add to [sfdnormalize](https://github.com/alerque/sfdnormalize)? Or are these legitimate changes, in which case is the normalizer's current enforcement of `SplineFontDB: 3.0` something I should bump to `SplineFontDB: 3.2`?

   *Edit:* Clearly this change is **wrong** because the fonts don't compile with this commit in place! But I don't understand why and it would be nice to know what's going on with FontForge.

2. Some of the font sources were encoded as `UnicodeBmp`, others as `UnicodeFull`. For the life of me I don't see any rhyme or reason why. I first got tripped up on this trying to grep for what fonts included a glyph with the intent of automating a change, but it threw me off to have the encoding different between fonts. @khaledhosny perhaps you can speak to this; besides the perhaps unnecessarily large encoding slot numbers in some cases is there any reason not to proceed with c009d21 and encode all the sources as `UnicodeFull`?

### DIFF
--- a/sources/LibertinusKeyboard-Regular.sfd
+++ b/sources/LibertinusKeyboard-Regular.sfd
@@ -48,7 +48,7 @@ MarkAttachClasses: 1
 DEI: 91125
 LangName: 1033 "" "" "" "" "" "" "" "" "Khaled Hosny" "Philipp H. Poll" "" "" "http://www.linuxlibertine.org" "This Font Software is licensed under the SIL Open Font License, Version 1.1." "https://scripts.sil.org/OFL"
 GaspTable: 3 8 2 17 1 65535 3 0
-Encoding: UnicodeBmp
+Encoding: UnicodeFull
 UnicodeInterp: none
 NameList: AGL For New Fonts
 BeginPrivate: 9
@@ -111,7 +111,7 @@ Grid
  432 575 l 1049
 432 580 m 1049
 EndSplineSet
-BeginChars: 65608 419
+BeginChars: 1114184 419
 
 StartChar: exclam
 Encoding: 33 33 0
@@ -11275,7 +11275,7 @@ Colour: ff00ff
 EndChar
 
 StartChar: .notdef
-Encoding: 65536 -1 295
+Encoding: 1114112 -1 295
 Width: 500
 Flags: MW
 LayerCount: 3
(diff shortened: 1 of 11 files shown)


khaledhosny:
> Commit [fca3f1e](https://github.com/alerque/libertinus/commit/fca3f1e5fc4424ca000e16f2d5a4ce4ebcb79b7e) is the result of opening every SFD file, saving it, closing it, and re-running the normalizer.

This seems to be broken, the glyph classes are truncated, some even mid-glyph:
```diff
- 78 parenright bracketright braceright backslash parenright.sc bracketright.sc braceright.sc
+ 78 parenright bracketright braceright backslash parenright.sc bracketright.sc bra
```

That `78` should be thee length of the line and FontForge uses it to decide who many chars to read (don’t ask me why it does not read tell the end of the line), but there are 88 chars in this line so FontForge is truncating the glyph classes and things went downhill from there.

> besides the perhaps unnecessarily large encoding slot numbers in some cases is there any reason not to proceed with [c009d21](https://github.com/alerque/libertinus/commit/c009d214b83e26470367cf6df3045ddb5aac3e10) and encode all the sources as `UnicodeFull`?

No there isn’t. You get lots of empty glyph slots when not using compact encoding (don’t also ask me why it is an encoding and not a view) and unencoded glyphs are way down, but that is not much bad from the status quo. But the only font that using non-BMP code points is the math font AFAIK.

khaledhosny:
> This seems to be broken, the glyph classes are truncated, some even mid-glyph:

This was broken in ab9bc72051e8eff031f39cb921afb146da02d6b4 which added `backslash` to some classes without also updating the number. This might be building fine because sfdLib ignores this number and reads to the end of the line (intentionally, I was bitten by this many times before).

alerque:
Thanks @khaledhosny (and sorry for accidentally editing your message instead of quote replying!).

I'll go back and fix what I broke, and then probably move ahead with `UnicodeFull` for all the sources just to ease any scripting I play around with.

PR #373, opened 2020-10-09
fix(keyboard): Correct dot placement to right of letter in ŀ U+0140

### LINKED ISSUE #372: Libertinus Keyboard glyph corresponding to [U+0140] is erroneous
This glyph corresponds to Unicode U+0140 code point "Latin Small Letter L with middle dot". The problem is that the dot should appear at the *right* side of lowercase "L", and in that font it is located to the left (before the "L")

In a Linux computer with X-Window system the «middle dot L» can be reached with Ctrl+Shift+U followed by the sequence "0140"+ENTER.

The companion glyph corresponding with the Capital "L" is correct.

**Screenshots / logs**

Here is a screen capture showing the problem with « ŀ » (in yellow), comparing this glyph in capital and lowercase form for 2 different typefaces: Libertinus Serif and Libertinus Keyboard.

image

alerque:
Closes #372.

@orestesmas You can test drive this from the artifacts posted in this PR under Checks→Fontship.

Are there any other placement issues with mid-dots like this in other font faces?


alerque:
It looks like the Mono face also had the backwards dot placement for U+0140. I've corrected it there too . I didn't check other middot glyphs, only this one, but this glyph at least is not the same across all faces.

orestesmas:
Thanks for the fast response.

Certainly, the Mono face had the same problem, I didn't notice.

As a side issue, I see that the "Display" face character [U+013F] (which is the Capital L with middle dot) has the dot located way higher than on the rest of the faces. I don't know if that is intended, or simply a bug. There are other glyphs with (in my opinion) misplaced dots in this face. For instance, the dots above the capital G [U+0120] and Z [U+017B] are also too high.

Do you think I should open a bug for that? The misplaced dot in " ŀ " was clearly a bug, but these ones are only aesthetic issues...

alerque:
Yes, plase feel free to open reports for visual issues that clearly don't match the aesthetics of other similar characters or the rest of the families. Obviously there is some degree of discretion in these but there are also patent inconsistencies that should be fixed and I don't mind making minor tweaks to make things more consistent. These sort of details matter in fonts!

It's helpful to have issues in bite sized bits of either the same issue/character across fonts or several similar character issues ... too many changes at once makes it hard to do and review, but obviously too granular is just noise.

PR #381, opened 2020-10-12
Fix dot alignment in capitals with "dot above"

alerque:
This is to address the side issue raised by @orestesmas in [comments on #373](https://github.com/alerque/libertinus/pull/373#issuecomment-706584550):

> There are other glyphs with (in my opinion) misplaced dots in this face. For instance, the dots above the capital G [U+0120] and Z [U+017B] are also too high.

While investigating this I looked up the Unicode spec to list all "with dot above" capitals. Most of the font families are okay, but Serif Display is a hilariously wavy line, and a number of other fonts have issues ranging from once glyph slightly off center to total chaos.

image

<details>
<summary>SILE Code</summary>

```sile
\begin{document}
\neverindent
\nofolios
\language[main=und]
\script[src=packages/grid]
\script[src=packages/rules]
\grid[spacing=35pt]
\grid:debug
\begin{raggedright}
\begin{script}

local files = {
	-- "LibertinusSerif-Regular.otf",
	-- "LibertinusSerif-Semibold.otf",
	-- "LibertinusSerif-Bold.otf",
	-- "LibertinusSerif-Italic.otf",
	-- "LibertinusSans-Regular.otf",
	"LibertinusSans-Bold.otf",
	-- "LibertinusSans-Italic.otf",
	-- "LibertinusMath-Regular.otf",
	"LibertinusMono-Regular.otf",
	-- "LibertinusKeyboard-Regular.otf",
	-- "LibertinusSerifInitials-Regular.otf",
}

	options = options or {}
		options.size = "8pt"
		SILE.call("font", options, { font, " — v7.010 vs. HEAD" })
		SILE.call("break")
		options.size = "30pt"
		options.filename = "/usr/share/fonts/OTF/"..font
		func(options, i, font)
		SILE.call("break")
		func(options, i, font)
		SILE.call("break")
	end
end

loop(function(options, i)
	local str = "ĊŻĖĠİȦḂḊḞḢṀṄṖṘṠṪẆẊẎȮ"
	-- U+010A U+017B U+0116 U+0120 U+0130 U+0226 U+1E02 U+1E0A U+1E1E U+1E22 U+1E40 U+1E44 U+1E56 U+1E58 U+1E60 U+1E6A U+1E86 U+1E8A U+1E8E U+022E
end)

\end{script}
\end{raggedright}
\end{document}
```

</details>


alerque:
@orestesmas It would be really great if you could test drive the fonts (which can be downloaded in the artifacts attached to the Fontship checks in this PR) with all these changes before I merge them. Not just the dot-above characters are affected, in some cases I moved the mark positioning anchors to be better centered, etc. This may have a trickle down effect on other unaccented glyphs.

alerque:
On hold until design choice about mark positions in #387 is resolved.

alerque:
The latest commit doesn't resolve #387, but it does put off the decision until late. For dot-above marks the primary change affects Serif Display. All others are pretty much just consistency issues, Display was the only one where the total average height got redrawn.

I've moved the combining mark up in Display so that the height is the same as the more sane glyphs before. This brings down the "high fliers" that were the original issue. It does still leave the Display face with much higher dot-above positions that any of the other faces, but that design decision can be put off until a full mark positioning overhaul is done.

New sample with Serif Display and Serif:

image

PR #383, opened 2020-10-12
Remove uni2098 and uni2099 from superscript kerning

### LINKED ISSUE #355: ₘ (U+2098) and ₙ (U+2099) subscripts (m and n) overlap integer subscripts
**Describe the bug**
When preceded by an integer subscript, subscripts `m` and `n` overlap the integer.

Subscripts `m` and `n` are represented respectively by ₘ (U+2098) and ₙ (U+2099)

This appears to only happen when preceded by an integer subscript, not another letter subscript. Furthermore, this only appears to affect `m` and `n`, but not other letter subscripts in the U+209x range.

**Steps to reproduce**
1. C₃ₘ
2. C₃ₙ

**Expected vs. actual behavior**
Expected: subscripts to render side-by-side
Actual: the `m` or `n` subscript overlaps the integer

**Screenshots / logs**

Bug:

Screen Shot 2020-08-25 at 3 36 35 PM
Screen Shot 2020-08-25 at 3 37 20 PM

But other U+209x work correctly:

Screen Shot 2020-08-25 at 3 38 00 PM
Screen Shot 2020-08-25 at 3 35 17 PM

Preceding with a letter subscript instead of integer subscript also works correctly:

Screen Shot 2020-08-25 at 3 41 06 PM

**Additional context**
This was observed with Sans Regular `.otf` file on both Android and iOS. Not sure if the platform matters, but since we've seen the issue on both I assume it's not too platform-specific.

clsn:
Closes #355.

Didn't try to do anything else, but fontforge may have done something.

Make of it what you will.  Hope it helps!


clsn:
Looks like letting FF mess with the classes wasn't so great; seems it has added some characters that aren't even there.  Sigh.

alerque:
Thanks for working on this. In spite of the mess you've probably saved me running down a lot of the same dark alleys. I'll have a look at this.

Whatever you did *did* fix this for Serif Regular. The rest of the fonts either don't build now or didn't get fixed. There are other issues here though, because a few fonts (even before this PR, so not your fault) have completely unrelated glyphs being drawn. For example the Display and Serif Semibold faces are drawing superscript circled plus and minus respectively instead of subscript m and n!

image

<details>
<summary>Details</summary>

```sile
\begin{document}
\neverindent
\nofolios
\language[main=und]
\script[src=packages/grid]
\grid[spacing=30pt]
% \grid:debug
\begin{raggedright}
\begin{script}

local files = {
	"LibertinusSerif-Bold.otf",
	"LibertinusSerif-Italic.otf",
	"LibertinusSans-Regular.otf",
	"LibertinusSans-Bold.otf",
	"LibertinusSans-Italic.otf",
	"LibertinusMath-Regular.otf",
	"LibertinusMono-Regular.otf",
	-- "LibertinusKeyboard-Regular.otf",
	-- "LibertinusSerifInitials-Regular.otf",
}

	options = options or {}
		options.size = "30pt"
		options.filename = "/usr/share/fonts/OTF/"..font
		func(options, i, font)
		SILE.call("kern", { width = "2em" })
		func(options, i, font)
		options.size = "10pt"
		SILE.call("kern", { width = "2em" })
		SILE.call("font", options, { "← "..font })
		SILE.call("break")
	end
end

loop(function(options, i)
	local str = "C₃ₘ C₃ₙ"
end)

\end{script}
\end{raggedright}
\end{document}
```

</details>

clsn:
I thought I saw a superscript circled plus at some point when trying these out, but couldn't find it again.  It would also maybe explain how we got into this mess: FF seems to think it knows what glyphs look like, or at least when they are similar to some others, and feels it has the right to tinker with kerning classes accordingly (this is what I noticed with changing the classes).  Perhaps it saw the superscript operators and figured, "Right, I'll class these with the other superscript things."  It feels like there may be some "ambiguity" of some kind, somehow two glyphs vying for the same codepoint.  I'll have to see if I can find that in the source.
Meanwhile... I guess I can try what I said at first, making the changes ONLY in the lines I meant to?  Whatever, see what happens.

waldyrious:
> these symbols are not encoded for math or chemical formulas and abusing them as such should be discouraged.

Tangentially related, but I thought I'd mention [this proposal](https://github.com/stevengj/subsuper-proposal) for adding mathematical subscripts/superscripts to Unicode — in particular, followers of this project may have interest in (and relevant input to) the discussion of [issue #7](https://github.com/stevengj/subsuper-proposal/issues/7) ("provision of fonts").

ygnessin:
Thank you so much @clsn and @alerque for looking into this. For someone like me who doesn't understand what most of tis font stuff means, can one of you clarify whether this is blocked or not and what the chances are that it will be fixed in the near term?

Thank you!!!

alerque:
@ygnessin I doubt we'll get all the issues with fonts that have different characters entirely in superscript/subscript positions sorted out for all weights any time soon, but for the specific case of `m` and `n` subscripts following digits that you brought up I expect we'll sort out the right kerning class to kill off the overlap issue pretty soon.

alerque:
@clsn I had edited the SFD files to *just* drop the two code points from the kerning class and nothing else and that seems to have done the trick by those two glyphs ₘ and ₙ. As noted in the related issue even though it didn't work I'd be interested in checking out the Python snippet you tried on this. Hand editing SFD files is only a couple grades up from writing assembly via punch cards.

@ygnessin Technical details aside, the upshot is that you should be able to give this a test run! You'll find development font builds in this PR under the "Checks" tab at the top. One of the automatic jobs that runs is called "Fontship". Just above the status report it should offer to let you download the artifacts from the test build. That should be a zip package (actually a zip in a zip) with fonts generated with these changes. Feel free to give them a spin and let us know whether or not they work for your purposes.

Note some of the faces still have bogus shapes encoded in these glyphs, but for the glyphs that are actually m's and n's they should be good to go. I currently get something like this:

image

alerque:
> Any kerning between subscripts and superscripts should be removed, these symbols are not encoded for math or chemical formulas and abusing them as such should be discouraged.

@khaledhosny I went ahead and did as you suggested and dropped the kern values for all subscript→superscript pairs as part of this PR for all font families except Math. Having looked at a bunch of sample cases and confirming it only *sort of works* half the time (only digits or narrow letters not wide letters, only for alternating one-each subscript/superscript combos, only if in the right order, etc.) I believe you are right and anybody writing stuff like this should use the Math font.

This might be the first truly breaking change I'm deliberately merging. I won't be surprised if we hear from folks that say they were using this 'feature' and it "worked just fine for them". I guess since I'm the maintainer now when the mobs come calling with pitchforks I can't throw you under the bus any more now can I?

khaledhosny:
> all font families except Math

Math does not it either, math layout engine align the scripts themselves and don’t need (and won’t use) kerning for that.

> This might be the first truly breaking change I'm deliberately merging.

I had the memory that I already dropped this kerning a while ago, but maybe I intended to but didn’t actually do it or didn’t do it properly.

khaledhosny:
>  won't be surprised if we hear from folks that say they were using this 'feature' and it "worked just fine for them".

It is really a bug, to spacing characters don’t get stacked on top of each other, not in Latin script anyway.

alerque:
> spacing characters don’t get stacked on top of each other, not in Latin script anyway.

Spacing characters?

The situation I was referring to that could be seen by some as a 'feature' regression is this (before/after):

image

It was possible to have both subscripts and superscripts before by stacking them in pairs. I agree this is an abuse and should be done other ways, but dropping the kerning is going to break any existing usages of such.

khaledhosny:
> Spacing characters?

These super and subscripts are spacing characters, not combining mark. They are encoded in Unicode as such and they work line a any standard base character, they just happen to be small and shifted up and down. http://www.unicode.org/versions/Unicode6.0.0/ch15.pdf#G12168

> but dropping the kerning is going to break any existing usages of such.

Which is a more reason to drop it. Any one depending on this is depending on a font hack to misrepresent data. That second rendering above is the correct, Unicode-compliant one. So it is a bug fix.

PR #405, opened 2020-11-17
Normalize sources, rebuild with updated tooling

alerque:
Round tripping through FontForge turned up some differences for me. I don't quite understand why.

More notably, the latest Fontship release that this will currently build on (v0.7.0) includes updated `fontmake` and `gftools` which handle a few things differently.


alerque:
I'm still confused about *why* some of these round trips introduced changes, but I've been consistently getting the same changes and have tested them using MWEs relevant to the code that gets changed and all seems to be well. Because it's really hard to review new PRs when all this stuff is getting mixed in every time I'm going to go ahead and merge.

alerque:
Okay I've concluded this has to do with my FontForge preferences now having "compact on open" enabled. The resulting resort of sources glyphs is catching some extant inconsistencies in the glyph numbering, hence most of these changes. They seem legit.

khaledhosny:
> I'm still confused about _why_ some of these round trips introduced changes

Someone added new glyphs by editing the SFD files manually and didn't set the encodings right.

PR #413, opened 2020-12-26
Twist Font Book's arm into classifying the Mono family

### LINKED ISSUE #411: Design style for Libertinus Mono not set to Monospaced on macOS?
In Font Book on macOS, there is Smart Collection that lists all installed fonts that have the Design style set to Monospaced.

Though I have installed Libertinus Mono on my Mac, but it isn't listed as Fixed Width font in Font Book, and it isn't available as an option in other apps that have list only monospaced fonts.

Libertinus Mono installed in Font Book

Libertinus Mono not listed in Fixed Width Smart Collection

Font version: 7.020
OS version: macOS Big Sur 11.1

alerque:
Closes #411

### DIFF
--- a/sources/LibertinusMono-Regular.sfd
+++ b/sources/LibertinusMono-Regular.sfd
@@ -25,6 +25,7 @@ TTFWeight: 400
 TTFWidth: 5
 LineGap: 0
 VLineGap: 0
+Panose: 2 0 5 9 0 0 0 0 0 0
 OS2TypoAscent: 894
 OS2TypoAOffset: 0
 OS2TypoDescent: -246


alerque:
I'm pretty sure this won't work yet, but @somelinguist can you check this build? You can find a ZIP as an artifact under the Checks tab of this PR. Documentation I found suggested forcing the Panose proportional=monospace flag here. Also as far as I've read this will **not** be sufficient to convince Font Book, but it's worth a shot before I try more aggressive hacks.

khaledhosny:
There is trick to satisfy the requirements of isFixedPitch while not breaking the combining marks, which is done by DejaVu Sans Mono: give mark glyphs the same advance width as the rest of the glyphs, then use a GPOS single positioning feature to zero their width (the last step was required for CoreText since it didn't zero the advance widths of the marks itself like Uniscribe and HarfBuzz does, not sure if this still the case).

khaledhosny:
Note that PANOSE monospace setting has the same issue as isFixedPitch, some implementations interpret it to literally mean all glyphs has the same width.

somelinguist:
I downloaded the artifact from https://github.com/alerque/libertinus/suites/1727574560/artifacts/32701674, uninstalled the previous version of Libertinus Mono and installed the version from the artifact, but it still doesn't show up as a fixed width font in Font Book or the apps that I tried. I'm not really as concerned about Font Book as hoping to use it in certain apps. I just assumed that those apps were using a list the the system provides. Thankfully, _some_ of those apps do give ways to choose non-fixed-width fonts even when they don't list them as such.

Thanks for work your work on this!

alerque:
As noted in discussion, this does not seem to be the right fix –or even any fix– to the problem. Since an actual fix will require a completely different direction I'm closing off this PR. Consider it a false start, lesson learned.

PR #419, opened 2021-01-03
Add slanted integrals

### LINKED ISSUE #357: Add slanted integrals
There is only upright integrals in this font. Could you please add slanted integrals as an option?

ivo-s:
Closes #357. The slanted versions were added using the `Italic...` function in FontForge, because it transformed the terminals more nicely than anything I tried by hand. Below are the shapes.

[integrals_shapes.pdf](https://github.com/alerque/libertinus/files/5761586/integrals_shapes.pdf)

An important note would be about italic correction and the placement of the limits, which I discussed in #357.

Like STIX, I created a stylistic set `ss08` to cover the slanted variant. I think it would be best to leave the upright version as default; I rather like it and it sits well with the font. I also added the display variants to the respective vertical-variants table. I learned some things that might be of interest for future reference.

When typesetting a display variant `$$\int$$`, TeX starts with an ordinary small-integral glyph. The stylistic alternative is applied first, and *then* the lookup for the display variant in the MATH table happens. When building the font, any `ss08` table entries regarding the `.size1` glyphs are ignored, because those glyphs are probably not recognized as used at the time (it's the bottom half of `integrals.fea`). Fortunately, this does not affect anything, as the vertical variants in the MATH table take care of the substitution. So far so good, just explaining this for the record.

As a suggestion for future revisions, the regular-sized path integrals have really small arrows. Because those glyphs are used in body text and are quite small, the direction of the arrow becomes illegible. Latin Modern solves this by disconnecting the ring. The reader then assumes that the speck on one of the ring ends is the arrow, and the direction of the integration is much clearer.
image

This PR builds on top of the added angle brackets, so I'll rebase it if needed after PR #417 is addressed.

### DIFF
--- a/sources/LibertinusMath-Regular.sfd
+++ b/sources/LibertinusMath-Regular.sfd
@@ -617,7 +617,7 @@ Grid
   Named: "Top Accent"
 EndSplineSet
 AnchorClass2: "aboveMark" "'mkmk' Mark to Mark-1" "right" "'mark' Right" "komb_OR" "'mark' Komb OR" "above" "'mark' Above" "below" "'mark' Below" "cedilla" "'mark' Cedilla" "ogonek" "'mark' Ogonek" "middle" "'mark' Middle"
-BeginChars: 1114819 4326
+BeginChars: 1114839 4346
 
 StartChar: exclam
 Encoding: 33 33 0
@@ -92391,7 +92391,6 @@ Width: 500
 GlyphClass: 2
 UnlinkRmOvrlpSave: 1
 Flags: MW
-ItalicCorrection: 150
 IsExtendedShape: 1
 GlyphVariantsVertical: uni222E uni222E.size1
 LayerCount: 2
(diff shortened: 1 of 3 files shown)


alerque:
Thanks for this! Mostly it looks good but there are a couple of bits that raise question marks in my mind (why did the italic correction value for unrelated glyphs change) so I'm going to have to give it a closer inspection when I'm prepared to actually test drive the results a little bit.

ivo-s:
Oh sorry I forgot to address that. It's connected to what I was discussing in #357. Italic correction determines the position of the limits. Since I was looking into it already, I corrected it where it needed correcting. To make sure that the limits' boxes don't collide with the integral, it is useful to look at something like `\int_W^H`. Some existing glyps needed a fix, so I did it. Shall I put together a PDF with old a and new limits (and perhaps all glyphs) to mark the difference?

alerque:
Thanks again. I've had a chance to review and I think everything looks good.

Issue #172, opened 2018-05-05
Merge Libertinus Serif and Libertinus Math sources


khaledhosny:
There are lots of duplicates between the two. Ideal we should have one source file and generate two fonts with different configurations (different line metrics, may be different defaults for some glyphs, etc.).

KrasnayaPloshchad:
This need some experiences on Python.

khaledhosny:
Note that I want to merge the sources, then decide at build time if we are generating the math or text font and subset away the unused glyphs.

jco2:
Okay I get it. I'm testing that on a new branch.

jco2:
I've noticed that the math font uses glyphs different than those in the serif font sometimes. So if we merge the sources we will have to choose which one we'd like to keep ?

To avoid that we could have one font with the only the math glyphs and merge the missing glyphs from the serif versions ?

KrasnayaPloshchad:
>I've noticed that the math font uses glyphs different than those in the serif font sometimes. So if we merge the sources we will have to choose which one we'd like to keep ?

We can comparing them to see which looks more identical to original Linux Libertine design, then keep that or improve.

khaledhosny:
Can you make a list and a visual diff between the two fonts so we can decide which is intentional and which should be normalized?

jco2:
The most striking differences between the two fonts are the brackets, braces and vertical bar. All the other glyphs in the math font look better.
I'm a bit lazy to make a visual diff though.
IMO we should keep the glyphs from the math font.

khaledhosny:
We want to keep the brackets, braces, vertical bar etc. from the text font, make the math version an alternate, and swap them based on the build target.

jco2:
Ok, so to do this I should start with the regular serif font, merge the math glyphs from the math font and move the math versions of the glyphs in a stylistic alternate ?

khaledhosny:
No need for stylistic set, just giving the glyph names a prefix, e.g. `.math`. May be better leave this merging of the fonts to me. It isn’t urgent and this issue is opened to remind me to do it at some point.

jco2:
Right, I'll work on the more urgent issue #63.

khaledhosny:
Not doing this.

alerque:
Re-opening this for consideration. See comments [here](https://github.com/alerque/libertinus/pull/423#issuecomment-758275705).

PR #423, opened 2021-01-10
Math: add some missing arrows

### LINKED ISSUE #403: Missing double arrows
Some double arrows like `U+21C4` ⇄ are missing in the Libertinus Math font. For a more complete list, see the following screenshot:

 screenshot of missing arrows in fontforge

This is in particular relevant for LaTex users, as `\leftrightarrow` does not work because of this.

ivo-s:
Closes #403. I filled in the block of arrows; here is a list of the new glyphs:

[new_arrows.pdf](https://github.com/alerque/libertinus/files/5792756/new_arrows.pdf)


alerque:
Awesome, great work as usual.

Can I ask when you post PDF previews you also include the Tex/SILE/ODT/whatever source code for generating them? These come in handy if I need to proof something or down the road need to fix a bug. I've been posting code snippets in my PRs & some issues using the collapsible content widget and so far I've gone back and used almost every one of them at some point. Someday I hope to have a more programmatic way to proof PRs both at submission and test for regressions later, but for now at if you at least post the sources you used that can save time reconstructing them later.

alerque:
Since most of these are not Math specific, we should probably copy them into some of the other (Roman style Regular weight) faces, no? I wonder if we can do that programmatically at build time so we don't have to maintain all the duplicates by hand. The Serif and Serif Math fonts have a lot of overlap and fixing things in both places is tedious and error prone.

ivo-s:
OK I recycled some code from back when I was changing a lot of math glyphs. You need to supply the .sfd and .otf files with the appropriate suffixes (OLD,REVISED), as per the strings in the Python and TeX file. I didn't include them because they are big. Then execute run.sh, which just checks for new glyphs using the Python script, and then renders the PDF using XeLaTeX.
[PR_arrows.zip](https://github.com/alerque/libertinus/files/5798072/PR_arrows.zip)

ivo-s:
I don't have any experience from the publishing industry, but I expected all these symbols to be relevant mainly for those oddball mathematicians who absolutely need a particular double-twisted arrow, because it signifies an essential algebraic relation :nerd_face: I would leave the decision to you; I don't know any typesetting needs outside of my field.

Regarding the optimization of the sources, @khaledhosny was considering consolidating the sources in #172. In general, I think it's a good idea, but some other people would need to be consulted as to what the best approach is. The ideal case, I understand, is to have a one big serif source file, which would simplify things by a great deal. The question then is, what the best practice would be in terms of splitting the glyphs between the serif and math fonts. TeX packages – I think – are OK with having a myriad fonts covering a single typeface, and even expect the text and math fonts to be different. But for many other uses, I imagine, people would just want one universal font file with as much Unicode covered as possible. On the other hand, this may not be very efficient if the whole font needs to be embedded somewhere (I only encountered this in Jupyter notebooks that have problems with some font encodings not permitting embedding of individual glyphs). Also there's the question of how to do it...prefixes in the sources were suggested, but maybe it's better to just leave it in some external table or config file, because FontForge expects the Unicode glyph names to be standard.

Finally, there are many math glyphs in the block starting at U+1D400 that actually come from other fonts like italic or sans, so there's always more opportunity for optimization :-) Although the whole no-kerning and italics-correction business is quite different in math mode, so that would need to be taken into account.

khaledhosny:
Re unification of the sources, there are some intentional differences between text and math fonts (like having an alternate glyph the default in the math font), and my idea was to identify these and have both glyphs in the source and switch between them at build time. The text font would also lack the math table and the math-specific layout features.

khaledhosny:
The math alphabets were originally taken from separate fonts, but I’d not split them again since even though the shape might be common, math spacing is quite different.

As for Unicode coverage, I’d let the regular text and math font have the exact same coverage, with the later having some different default glyphs and math layout.

alerque:
Thanks again for the contribution. Because adding these to the regular Serif family would introduce the problem of not having other weights and styles for them I think I'll go ahead with just adding them to the Math family for now. Merging sources can be a topic for another issue, as could adding other these to other families or weights.

alerque (approved):

PR #431, opened 2021-01-17
Create libertinus.css

Crissov:
shows how to use Libertinus from CSS, including stylistic sets


alerque:
How was this file generated? This looks like something that should be baked into the build process to be packaged in releases more than a source we need to maintain in the repository. As such I'm guessing it should go into Fontship, a feature idea already tracked in [issue 77](https://github.com/theleagueof/fontship/issues/77).

Crissov:
I've written most of the CSS by hand, looking at the `.fea` files. I believe it is mostly documentation, but could perhaps be automated to some extent.

alerque:
Hmm, okay. That seems like a lot of work given that I've seen tooling which pretty much discovers and writes all this boilerplate automatically. I've also seen several other projects including CSS files but they don't seem to be very normalized, i.e. just because you setup a site with one font doesn't mean you can swap out another well-configured one by exchanging CSS sources.

I'm wondering what's the best step forward. It seems like if we're going to include something like this by default we should make sure it is complete and stays up to date (i.e. automatically generated if possible) and is interchangeable as possible (i.e. sites with basic consistent styling should be able to switch fonts with similar feature sets). Do you know what the scene for this is? Or am I dreaming just to think those goals are even achievable?

alerque:
Hey @Crissov can you add a commit (to either this or your other PR) that adds your name to the AUTHORS & CONTRIBUTORS files?

alerque (approved):

PR #436, opened 2021-01-30
Remove Spiro curves to fix build issues

ivo-s:
I noticed a buggy glyph `uni2639` in Serif Italic, Sans Regular, and Sans Italic fonts. The source of these glyphs uses Spiro curves, which are supported by FontForge, but the building process seems to have some problems dealing with them. I removed the Spiros as the easiest solution, and there should be none left in the sources.


khaledhosny:
What is the exact issue? If [sfdLib](https://github.com/alif-type/sfdLib) is chocking on them, it should be fixed (by ignoring spiro sections, of course).

ivo-s:
I don't know, becuase I am not that familiar with the building process. The built glyphs look like 4 of the 5 paths were joined together:
image
In the sources, it seems there are both Spiro curves and Bézier curves. FontForge displays only one of them, depending on the InSpiro glyph variable. In any way, I think having both is redundant.

Of course, it would be good to find out if there is any problem with the building tools. In this regard, I was counting on you guys, because you have the knowledge :-) However I can try to track it down, it's just going to take some time.

khaledhosny:
This is now fixed in sfdLib (I tagged v1.0.7 as well).

alerque:
Thanks @khaledhosny for the fix in sfdLib. The dependency has been bumped in Fontship and (as soon as Docker catches up) builds for this repo should start using it too.

That being said, if I understand what @ivo-s is saying and they aren't serving any purpose here it seems like it might be a good idea to merge this PR anyway — not because they break anything just to cleanup unused bits. Thoughts?

khaledhosny:
Apparently whoever draw these glyphs used Spiros, but FontForge with always save Béziers next to them, so it is up to you whether the Spiros should be kept as the source or not.

waldyrious:
> Apparently whoever draw these glyphs used Spiros

That would be me (#307) but I don't have a strong preference to use Spiro curves — IIRC, I just found them neater than Bézier for representing the shapes of those characters. Feel free to adjust the glyphs as you guys see fit.

alerque:
I think the verdict for now is that we'll keep the Spiro curves in the sources. If they were more intuitive for drawing they are probably more intuitive for editing as well and with the build issues fixed I don't see any harm in keeping the source data.

PR #440, opened 2021-02-17
Fix Q.u bearing

ivo-s:
This tiny fix concerns the `Q.u` glyph in the sans font. It should have the same bearings as `Q`, as it does in Serif Regular and other fonts. It is necessary for proper kerning.
Currently (top), the "Qu" pair is too close. Fix (middle); ordinary Q glyph (bottom, kerning is zero for this pair).

image


alerque:
Are any of the other sans styles affected by this?

alerque (dismissed):

ivo-s:
No. I checked the other Sans and Serif fonts, both caps and small caps versions of this. I found about two other anomalies where there is a tiny 1-to-3-point shift, which is not really visible, so I left them.

alerque:
Lets go ahead and normalize these so they are all the same. 3 units isn't really discernible side by side, but even 1 unit is noticeable if you flip feature flags on and off and watch your document re-render. I noticed in LibreOffice testing just now because 1 unit can cause big changes in antialiasing. Ironically this doesn't show up so much in the `u` but it does show up with other letters farther down the line! 1 unit even has the potential to cause reflows. I think as long as we are fixing the visible problem we might as well actually do it right.

ivo-s:
Alright, I went through all `Q.u` and `q.sc.u` glyphs and fixed them. In some cases, the point positions were rounded to whole integers, while the small-cap letters were rounded to `0.1`. In one case, the thickness of the glyph near the bottom was different, probably because some Bezier handles did not reflect a different position of the Q tail. Basically, the extended Q glyphs should be just like an O with a tail, as they are in the rest of the files. In Serif-Italic, the `q.sc` glyph bearings were fixed, because they were clearly off with respect to similar small-cap glyphs.

The changes are really tiny and they can be seen when one opens two characters in FontForge (mouse double-click, otherwise the viewport won't be the same) and switching between tabs. The corresponding pairs are `Q - Q.u`, `q.sc - q.sc.u`. All these pairs should now be completely identical escept for the tails.

alerque (approved):
Thanks for the contribution. I've just rewritten the commit messages and added a changelog entry.

PR #443, opened 2021-02-22
Dotless vs. dotted i/I in KAZ and TAT

Crissov:
The font feature files were already handling the roman orthographies of the Turkic languages Azeri `AZE`, Crimean Tatar `CRT` and Turkish `TRK`, but neither Tatar `TAT` nor Kazakh `KAZ`. (The official roman orthography for Kazakh has come into effect only very recently.)


sources/features/sc.fea
```diff
@@ -245,12 +245,19 @@ feature smcp {
     language dflt;
       lookup smcp_smallcap_i;
       lookup smcp_smallcap;
+    # exempt Azeri, Kazakh, (Crimean) Tatar and Turkish:
```
alerque, line 248: Lets drop the comment. Sure commenting is a good practice in general, but nothing even much more obtuse is currently documented in the fea files and this particular case is pretty obvious (and somewhat ironically as a fluent Turkish speaker married to a Kazakh I'm pretty familiar with why this exception exists). Meanwhile I've been playing around with auto generating the FEA files (from FEE) so the less fluff there is to compare with the easier it will be to eventually validate the switch.

alerque (dismissed):
Thanks for these fixes!

alerque (approved):

PR #444, opened 2021-02-22
documentation for font features

Crissov:


alerque:
Thanks for the contribution. Can you tell me how this was created/generated? It is more complete and in a format I'd prefer that the existing ODT based documentation. Obviously one is documenting the FEA source and the other more the OpenType feature end of things, but there is also substantial overlap. Ideally I'd like to settle on who we are documenting for and in what format and bring these together a bit.

sources/features/readme.md
```diff
@@ -0,0 +1,151 @@
...
+form default ligatures.
+
+The related feature `clig` is not used.
+
+The respective CSS property is `font-variant-ligatures` with the value `common-ligatures` (or `no-common-ligatures`).
+
+Discretionary and Historical Ligatures: `+dlig`, `+hlig`
+--------------------------------------------------------
+
+The only historical ligatures supported are Latin lowercase _ st_ and _ct_ in all faces.
```
waldyrious, line 44: Typo, I suppose:
```suggestion
The only historical ligatures supported are Latin lowercase _st_ and _ct_ in all faces.
```

Crissov:
JFTR, the markdown was taken from the documentation I did for the CSS file in #431. Unfortunately, there's nothing generated automatically yet.

alerque:
Docs are kind of a disaster right now. We have a preview image that needs LaTeX to regenerate, LibreOffice sources with generated PDFs stored in Git, and several assorted Markdown files.

While I love markdown and *could* setup conversion to PDF as part of the build release process, there are two things that don't play very nicely yet. 1 is math, 2 is "live" previews of the font and features being documented using the actual font. Both can be handled but it's a complicated mess.

However I think all that's about to change. With math support [about to land in SILE](https://github.com/sile-typesetter/sile/pull/1103) we'll have access to a relatively lightweight way to keep sources in a text based format and have math and previews of the current font baked into PDFs with each rebuild.

All that to say I think I'm going to go ahead and merge this now and work on merging the different documentation bits and massaging them into a better format for distribution later.

alerque (dismissed):

alerque (approved):

Issue #345, opened 2020-08-18
Design inconsistencies in the math font


philosaurus:
The design of the math font is inconsistent. Sometimes this is just a minor thing not visible unless the glyphs are carefully observed under magnification, but in a few cases (points 1–2 below) the design feels simply wrong.

1) U+2200, U+2201, U+2203 and U+2204 look like they're from another font. They are way thicker than other symbols (60+ em vs the regular 50) and their style is completely different to the rest.

2) U+221F, U+2227, U+2228, U+2229, U+222A, U+228C, U+228D and U+228E are visibly smaller than the rest of the operators, and, most importantly, are not centered on the math axis. U+2229 and U+222A should probably just be rotated versions of U+2282, but even with the current, narrow design the placement is definitely wrong.

3) Some "barlike" operators are too thin (44 em vs the regular 50): U+007C, U+2215, U+2216, U+2223, U+2224, U+2225, U+2226.

4) Most operators seem to have have cupped line endings, but some have sharp corners. (Maybe they are earlier designs?) This is just a minor consistency issue. Affected are (at least): all operators listed in the previous points, U+2205–U+2207, U+2211 (compare to U+2210), U+221A–U+221C, U+2220, U+2221, U+2222, U+223E, U+223F, U+22A2–U+22AF, U+25B7, U+25C1, U+266F, U+27C2, U+27F2, U+27F3.

5) The negation slash in U+2209 and U+220C is 50 em thick, but for the relational operators it is only 41 em.

alerque:
I'll evaluate these at some point (not super soon), and additionally I'm getting closer to being able to review and potentially accept community contributions too if somebody else wants to tackle these.

But...

First please break this up into separate issues. There ar at least 5 maybe more issues that could be dealt with separately here, some maybe acted on, others not, some maybe needing some iterations & design review. That gets very messy to track in a single issue. Please split this up with one issue per design feature that would be a discrete decision to make and/or task to be reviewed in a single PR with a single set of comparison images.

philosaurus:
I didn't want to spam the issue tracker, but I'll break it up as per your request.

alerque:
Closing in favor of itemized issue reports #346, #367, (to be continued?)...

philosaurus:
I've moved all others to their own topics, left one of the issues in here.

ivo-s:
Hey @philosaurus, see #454 concerning this issue. I left some glyphs unchanged; discussion is welcome.

PR #454, opened 2021-03-07
Math style fixes

### LINKED ISSUE #346: Some mathematical glyphs are out of style
U+2200, U+2201, U+2203 and U+2204 look like they're from another font. They are way thicker than other symbols (60+ em vs the regular 50) and their style is completely different to the rest.

libertinus-math-out-of-style-glyphs

### LINKED ISSUE #347: Some operators have wrong sizing and alignment
U+221F, U+2227–U+222A and U+228C–U+228E are placed on the baseline rather than being centered on the math axis. They also seem to be a bit small. The U+2229 and U+222A should probably be more similar in appearance to a rotated U+2282.

The + and × are only included in the image for reference.
operators_wrong_size_placement

### LINKED ISSUE #348: Inconsistent line thickness in math font
1. Some "barlike" operators are too thin (44 em vs the regular 50): U+007C, U+2215, U+2216, U+2223, U+2224, U+2225, U+2226.

2. The negation slash in U+2209 and U+220C is 50 em thick, but for the relational operators it is only 41 em.

### LINKED ISSUE #406: Style of math operators distinctively changed after PR ivo-s:math-fix
The opening of this issue comes from a brief discussion in the comments of the pull request ivo-s:math-fix adressing the issues #63, #164, and #227. The PR changed the style of many of the glyphs besides fixing the size and alignment problems. 
This [file](https://github.com/libertinus-fonts/libertinus/files/3099005/operators_compare.pdf) from the PR post compares the glyphs side by side. 
image
The style of some characters is distinctively changed. For example, the symbol ± is now connected and the aperture of characters such as ≤, ⊂ is wider.

ivo-s:
There is a number of open issues relating to mathematical glyphs, mainly addressing style inconsistencies. Some are oversights, some are orphans from earlier versions. The overall number of mathematical glyphs is overwhelming, so I follow a per-need basis. List of changed glyphs:
[LibertinusMath_compare.pdf](https://github.com/alerque/libertinus/files/6097149/LibertinusMath_compare.pdf)

## Resolved issues

* Fixes #406. I decoupled the ± glyph. The upper arm is a bit longer for optical symmetry. The minus lies on the baseline, not below, because this glyph is more often used with numbers and in-line text, as compared to, for example, ≤. (same choice in STIX) The set operators were also changed, see below. I decided not to narrow the <,> glyphs, because the angle of 52° and the size is quite standard for a math font. I would not see the original narrow glyphs as a design choice, because the mathematical glyphs in Linux Libertine were not designed for a mathematical font that needs big operators. In my opinion, the current version offers optimal legibility and can be more easily distinguished from ≺, ≻.
* Fixes #348. The vertical bar is now thicker, 50 points, and has 80-80 bearings (previously 81-80). I mainly encounter this glyph as a bracket and it definitely looks better in this role. I didn't change anything in the text fonts, which also holds for other mathematical glyphs. For the time being, I see it as a safe choice. I didn't touch the division and fraction slashes, U+2215  ∕, U+2044 ⁄. The reason is that they seem to be tied to other Unicode glyphs like fractions and units. I never see them as standalone glyphs. In math, there is always a normal solidus U+002F used for division, which has a standard thickness 50.
* Fixes #347. Here I went to other fonts for guidance, because from my hand-writing and school experience, the logical and set operators were always put on the baseline. It seems that both conditions need to be met - the math axis and the baseline. Becuase the set operators ∩, ⊂, should be just rotated versions, this imposes a condition for square symmetry. The operators are now consistent. I also revised my initial decision to compose them of a circular arc and two straight lines. Now, they are more pleasantly rounded in my opinion.
* Fixes #346. Fixed height, thickness, and style.

## Issue #345 discussion

Some of the glyphs from #345 have been addressed, but I am not convinced about all of them. I don't think that all caps should necessarily be rounded. I am not familiar with the symbolics of some glyphs like ∾; some like delta and nabla should _not_ have rounded edges, because neither does the corresponding Greek delta; some symbols like geometric shapes or musical sharp sign are not mathematical symbols. The sigma sum could have rounded edges like the Greek sigma, but the person who drew and scaled the bigger version of the sigma and pi operators U+220F, U+2211, clearly knew what they were doing, and I didn't want to disrupt this operator set. We can discuss this one.
This is why I haven't marked #345 as resolved.

## Operator alignment

In #347, the issue of operator alignment was referenced. The idea is that the vertical position of the operators should follow the numbers and the features `lnum, onum`. This would be neat in some situations, but generally I would advise against it. The main reason is different usage (at least in my experience, which is math and physics). I have never seen any number-related OTF features in math mode. The operators are centered on a math axis because, in most math in my experience, the operators are put between letters, not numbers. The math axis is a compromise that should work for both lowercase and uppercase letters. Old-style numbers would basically require a separate axis corresponding to lowercase letters. As far as I know, old-style numerals are never used in math. This whole feature would also most likely require duplicating all operators just for this feature. That is why I don't think it's worth it.

I will ping the individual issues to allow some discussion. I am also going to count on @alerque to handle the changelog and commit squashing upon merge.


ivo-s:
I thought I already nailed it, but now that I'm looking at the PDF, I am thinking whether the arc in uni2221 and uni2222 should not extend more beyond the wedge. What do you guys think?

philosaurus:
Having only taken a look at the attached PDF, the changes seem to be for the better :)
Looking forward to seeing them in a live document.

thlinard:
> In #347, the issue of operator alignment was referenced. The idea is that the vertical position of the operators should follow the numbers and the features lnum, onum. This would be neat in some situations, but generally I would advise against it. The main reason is different usage (at least in my experience, which is math and physics). I have never seen any number-related OTF features in math mode. The operators are centered on a math axis because, in most math in my experience, the operators are put between letters, not numbers. The math axis is a compromise that should work for both lowercase and uppercase letters. Old-style numbers would basically require a separate axis corresponding to lowercase letters. As far as I know, old-style numerals are never used in math. This whole feature would also most likely require duplicating all operators just for this feature. That is why I don't think it's worth it.

As I introduced the idea, let me clarify it.

Libertinus Serif has a `case` feature (we could wish for a few more glyphs, but that's already great). This allows for example (the third hyphen is a `hyphen.case` glyph):
Libetinus Serif case

I think a math axis that would be a compromise between the uppercase and lowercase (oldstyle) figures would be poorly designed. And I speak voluntarily of uppercase and lowercase figures, because in my opinion they shouldn't be seen otherwise, and the terms of "standard figures" and "old-style figures", although more conventional, obscure what should be their correct perception and treatment. In fact, I firmly believe that there should be as many math axes as there are figures cases, i.e. one or two for most fonts: if there are only uppercase figures or only lowercase figures, a single math axis, and if there are both, two math axes.

Libertinus Math only has uppercase figures, and, fortunately, its mathematical axis is not a compromise, but it's that of uppercase figures **and uppercase letters** (I'll come back to this).

So my remarks would be more directed towards Libertinus Serif (and I saw that wasn't the subject of this thread, but let's talk about it anyway). Libertinus Serif has lowercase figures, and the math axis is aligned to the axis of "letter operators", like hyphen:
Libertinus Serif axis

Where things go wrong is when you put the text in all caps (InDesign has a feature like this, it change the case and activate `lnum` and `case` features): the figures follow (they become uppercase), the hyphen too, but not the math operators:
Libertinus Serif uppercase1

While it should be like this:
Libertinus Serif uppercase2

And coming back to Libertinus Math, it's actually well designed on this point: its math axis couldn't be much different from the uppercase letter axis (in the following example everything is in Libertinus Math except the hyphen, which is the hyphen.case of Libertinus Serif):
Libertinus Math uppercase

Hope it's clearer like this. I'm not arguing for lowercase figures to be added to Libertinus Math. But if this were to happen, or for fonts that have two figure cases (like Libertinus Serif), you need to take care of the two math axes: one in the same vertical axis as hyphen, dash, etc., and the other in the same vertical axis as `hyphen.case`, `dash.case`, etc.

khaledhosny:
> I decoupled the ± glyph. The upper arm is a bit longer for optical symmetry.

I find the old glyph to be optically balanced than the new one. The old glyph is also closer to the Computer Modern one. In STIX Two math font, the plus and minus are not attached, but the lower arm is the longer one.

ivo-s:
@thlinard I see that you are arguing the principle, not an implementation. From typographical point of view, I agree with your examples of uppercase/lowercase figures, and that operators should ideally follow the height of the glyphs that surround them. However, the problem to which I was referring, is that only one math axis is permitted by the OpenType format. As far as I know, this parameter is used to draw fraction lines so that fractions align with common horizontal operators. If one shifted all the operators, the elements that depend on the math axis parameter – like fractions – would be misaligned. For this reason, a suitable compromise needs to be found, because mathematical formulae use both uppercase and lowercase glyphs. Perhaps it's poor design, but it is a necessity.
I noticed that in your last picture, the math operators are placed very high and do not correpond to the math axis of the font. Is there something else at work? My math renditions look like this:
image

@khaledhosny I don't have any strong preference on this. In #406, it was argued that some of the design choices from Linux Libertine should be maintained so that that the typeface does not imitate Computer Modern too much. I think I read @alerque's comment somewhere where he was in favour of splitting the ±, but I can't find it.
STIX does have a variant where the length ratios are different, but contrary to CM, the plus does not align with the + glyph. Also it seems to me that it makes the plus look uneven, but it's a subjective impression.
Let's first decide on which variant is better, and if we pick the decoupled glyph, I'm more than happy to get advice on good optical balance. We can also put the minus below the baseline.

thlinard:
> I noticed that in your last picture, the math operators are placed very high and do not correpond to the math axis of the font. Is there something else at work? My math renditions look like this:

You're right, my bad. This is the correct example:
Libertinus Math corrected

For Libertinus Math, I agree the problem is probably insoluble. The chosen math axis doesn't really work with lowercase surroundings (too high) and doesn't work neither with uppercase surroundings (too low), but it does indeed seem difficult to do otherwise. It must be said that Libertinus has one of the smallest x-height (with Latin Modern and Garamond Math) among the math fonts that I know and that doesn't help: with less contrast between the lowercase height and uppercase height, the compromise would be less visible.

On the other hand, for Libertinus Serif (or Libertinus Sans, for that matter), we can very well consider two mathematical axes. Figures and math symbols in Libertinus Serif don't serve quite the same purpose as in Libertinus Math: rather for short, much simpler formulas (and besides the operators aren't the same size). The problem wouldn't arise if there were only uppercase figures, but since there are both, it would be consistent if there were two sets of operators.

The `onum` feature in Libertinus Serif contains only `0123456789`. In Stix Two Text, it contains `0123456789$€£`, and `0123456789$€£¥₹₺₽` in Cambria. We could imagine a few more symbols, closely related to figures (`%‰°`), but above all an alternate set of mathematical operators as found in common charsets (`~<>=×÷+−¬±∞≈≠≤≥◊`).

cionx:
Regarding the changed length of “equality-like” symbols such as `⊆`:
Maybe it makes sense to have these symbols all have the same consistent length (i.e. the length of the equality sign). This should (hopefully) ensure that when these symbols are aligned underneath each other, the contents next to them are aligned as well.

This PR is already going more in this direction, but is still a tiny bit off.

<details>
<summary>current version</summary>

current

</details>
<details>
<summary>new version (from this PR)</summary>

new

</details>

ivo-s:
@cionx I am open to changes, but that would be too many conditions on the set operators :-) See the third point in my initial post – I believe that the set operators should be on the baseline, but also centred around the math axis. This creates a unique condition on the height of ⊂. Additionally, other set operators like ∩ should be just rotated (and on the baseline and around math axis). This creates a condition on square symmetry and makes it impossible to scale the width to match the equals sign. I think that the difference is tiny and the alignment can be handled by the typesetting engine. Also in my experience, one usually needs to align equals/less/more symbols rather than a mixture of arithmetic and set operators.

Having revisited this PR, I hope I can summarize the remaining open questions.
* The ± glyph. Should it be joined or not, and how the optical balancing should be done. Accepting any advice (or decisions from Caleb).
* uni2221 and uni2222, maybe the arc should extend a little bit more.
* uniform width of relational operators. I generally honor this principle, but I believe the set operators should be an exception for the reasons above.

ivo-s:
I decided to modify the plusminus a bit; I don't think it has to align with the plus, these two symbols will never be side-by-side anyway. I think it now looks good in context:
image
The symbols uni2221 and uni2222 now have the circular arcs extending a bit more, which looks better in smaller font sizes.
Up-to-date changelog:
[LibertinusMath_compare.pdf](https://github.com/alerque/libertinus/files/6979151/LibertinusMath_compare.pdf)

ivo-s:
@alerque I do not plan on making any further changes, as I think everything has been resolved.

Firestar-Reimu:
@ivo-s anyway to fix this issue? https://github.com/alerque/libertinus/issues/464

ivo-s:
> @ivo-s anyway to fix this issue? #464

That is a big one and not really the subject here. I gave it a few hours here and there, but I most likely won't have time to tackle it in the near future. If I remember correctly, there are compromises to be made between italic-style font bearings for characters that follow each other to look natural, and italic correction that both symmetrizes the character in upright context and influences sub- and superscripts. Cannot tune all three independently.

alerque (dismissed):

alerque (approved):

PR #460, opened 2021-03-09
Dutch ij digraphs with acute

### LINKED ISSUE #437: Dutch-specific replacement of IJ
Heijn

Language-specific OTF features are active in Libertinus and work as far as I tested*.
However, in Dutch, there is a digraph/letter/ligature consisting of I and J. There are Unicode points defined for both the upper and lower case form (U+0132/-3, `Ĳ` and `ĳ`) that look (as far as I can tell) identical in the upright form in Libertinus, i.e. it makes visually no difference entering `IJ` or `Ĳ` (`ij` or `ĳ`) in XeLaTeX.
**Problem:** The italic Libertinus fonts display U+0132/-3 differently than separate I and J as can be seen in the image above.
**Expected:** When the text is marked Dutch, the italic font should look like the second forms regardless whether `IJ` or `Ĳ` (`ij` or `ĳ`) is used (i.e. one should not have to resort to enter U+0132/-3 in ones text to get a pleasing output). As the IJ is common in Dutch, it really should look perfectly of the box.

It might be worth mentioning that there is no combined version (like Ij) since at the beginning of a word, *both* letters are set upper case: [IJssel](https://en.wikipedia.org/wiki/IJssel).

There are (rare) words that have a coincidental i+j clash such as *bijectie* (from *bi-* and *ject*, Latin for "throw") that one would have to handle manually, but that's a general issue with ligatures, end forms and similar features.

Furthermore, IJ/ij can carry an accent (mostly the acute) and when it does, ideally, [both letters should carry it](https://en.wikipedia.org/wiki/IJ_(digraph)#Stress).
Upright IJ/ij with any variant of placing an accent on the letters I and J separately looks as expected. Using U+0132/-3, however, looks terrible. (This state is acceptable, virtually no one uses U+0132/-3 when typing text.)
Italic IJ/ij/U+0132/-3 look all terrible with an accent on them. (The fact that separate I and J with accent look bad is purely due to J/j not working with accents; that italic U+0132/-3 don't work with accents is acceptable.)

*) I tested: Turkish small caps (keep the dot on a small i) and Serbian italic Cyrillic (*б г д п т* shaped as expected).

ivo-s:
Fixes #437.
This change adds the ij digraph as a `locl` feature for Dutch (NLD), but not Flemish, as the treatment of ij seems different there:
https://en.wikipedia.org/wiki/IJ_(digraph)#Flanders
...aaand it works with acute! :partying_face:  Behold:
image
[ij_test.zip](https://github.com/alerque/libertinus/files/6111053/ij_test.zip)
As long as this PR is active, you can download the compiled font files from the forked branch and test them with the source above:
https://github.com/ivo-s/libertinus/runs/2069981150

The substitution should work for uppercase, lowercase, and smallcap letters. Combining acute works with `íj`, where the `í` can be a Unicode glyph or i with a combining acute. Considering the limited font support of acuted ij, I imagine this is how most people would write it anyway. The substitution also works with the diagraph `ij` plus combining acute. All possibilities are laid out in the testing XeLaTeX source, which is encoded in UTF-8.

You may notice that Semibold Italic does not have a ligature between the lowercase ij, even though Italic and Bold Italic do. This can be fixed in a future update, preferably by a type designer who can draw nice curves :-)

I also swapped `acute.cap` and `grave.cap` in Serif Semibold Italic, because I happened to notice they were switched.

I corrected the small caps feature `smcp` behavior, because it did not work for some languages. This is due to the feature file specification:
https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#4.b.ii

> If your font has several languages for a given script, and you need language specific rules for only some of the languages, you should still explicitly name all of the languages so that they will inherit the script-level default rules.

That is also why I added a reminder comment in `gsub.fea`.

I expect that we should also decide whether to make a separate i_j glyph, as suggested by @georgd.


ivo-s:
I added the *I_J* and *i_j* glyphs for substitutions and I also put *I_J* and *IJacute* in some kerning classes that already contained *IJ*.

alerque:
Thanks for the contribution! Sorry it took me so long to review.

alerque (approved):

alerque:
I rewrote the commit messages (I'm still playing around with what commit message template I like for font projects!), added a fontlog entry, and merged some changes so that it works with Fontship v0.8.0 (which has a new version of sfdLib with API changes) so that the CI checks pass. I haven't dug in too deep on the alternative implementations but as far as I can tell this fits the bill.

Bolpat:
> but not Flemish

That's a mistake. What matters isn't what school teachers say, but how the digraph is perceived. I have a custom stylesheet for Wikipedia and, well, have a look:
Flanders

PR #463, opened 2021-03-13
Clean up and modernise Serif sample document.

nsemrau:
- Change the compared font of Times New Roman to its metrically
  compatible free-as-in-freedom equivalent Liberation Serif

- Change the manually inserted per-page header into an actual header
  - Remove one sentence at the end of a longer paragraph per row. It
    keeps the old structure of page two and it thusly does not overflow
    into page three due to the new header

- German text:
  - Call it "page" instead of the German "Seite" in the header
  - Remove the German translations of the header and the
    footnote
  - Fix a missed space in the German text and the header

- Fix Linux Libertine upstream bug 088 "Wrong icelandic text in
  multilingual.pdf" (https://sourceforge.net/p/linuxlibertine/bugs/088/)
  while we're at it

- Recreate PDF version with above changes
  - Use Libertinus 7.040 instead of 6.4 to render the font

The German text seemed superfluous since Libertinus is a Github-driven project in English. The document was a mish-mash of German and English, so I decided to normalize that aspect. I'll re-add the translated footnote and header if you'd like to keep it.


fitojb:
I’ve just noticed another (pre-existing) misspelling: “Deos” should be “Deus” in the Portuguese sample.

I just mention it since it might save time to fix it at the same time as the other typos.

alerque:
Thank for you this contribution. Eventually I'd like to ditch the PDF from the repository entirely and regenerate it in CI on PRs and releases. I'm going to merge this because it's an improvement and I believe in iteration, but if we have to touch it again lets write a SILE specimen document that is easier to render on the fly.

alerque:
This will be replaced in the next release with a SILE rendered document that updates on each build. Style and content contributions towards samples still welcome. See #401 for where the new sample page is landing.

khaledhosny:
IMHO, comparing with Liberation Serif does not have mych value. It was comparing with Times New Roman because Linux Libertine was designed as a replacement for it, but other than metric comparability, Liberation Serif has no resemblance to Times New Roman. Using a libre Times clone like TeX Gyre Terms might be a better alternative.

alerque:
> Using a libre Times clone like TeX Gyre Terms might be a better alternative.

I agree, I thought the same thing while working on re-typesetting that. I also need to get whatever font is used installed in CI, because right now I'm not even sure it is being rendered as what it says on the tin :wink:

PR #471, opened 2021-04-17
Adapt for Fontship v0.8.0

alerque:


tools/build.py
```diff
@@ -104,7 +103,7 @@ def _post_process(self, otf):
         gdef = otf["GDEF"].table
         classdef = gdef.GlyphClassDef.classDefs
         for glyph in font:
-            if glyph.lib.get(GLYPHCLASS_KEY) == "mark":
+            if glyph.lib.get(CATEGORIES_KEY) == "mark":
```
khaledhosny, line 106: FWIW, with latest sfdLib and latest ufo2ft (not released yet I think), you will not need any of this dance and GDEF glyph classes will be handled out of box (also the version of sfdLib you are using is slightly broken, categories should be in Font not Glyph lib).

alerque: Yes I had a quick look and it seems like most of this script will be obsolete soon. I'll get the rest that isn't specific to this project moved to Fontship eventually, this was just to tide things over so CI wasn't broken because I pulled a newer sfdLib into Fontship (which doesn't have a good way to hold back dependency versions).

PR #476, opened 2021-05-20
Fix Finnish translation of In the beginning...

NikoDaGreat:
In the Finnish translation of *In the beginning was the word*, there is a typo, 

    hä -> hän

The translation is a bit odd overall, but at least the typo should be fixed and the *ja* capitalised.


alerque (approved):
Fair enough ;-) Thanks for taking the time to contribute.

Issue #340, opened 2020-08-08
Hebrew diacritics positioning


KrasnayaPloshchad:
[Here](https://github.com/alerque/libertinus/issues?q=label%3A%22script+hebrew%22+is%3Aclosed) are series of bug related to Hebrew diacritics, the previous maintainer have no ideas on them, and finally closed them.
Now I’ve found there are some good reference implementations on them, provided by Culmus project.
http://culmus.sourceforge.net/taamim/index.html
This page give some Hebrew fonts with good mark attachments, available with both TTF font file and SFD source, so every developer would possible to find a solution to fix. So I leave this resource here and everyone who has interests or idea in such bug fixes would get benifits.

reimerd:
The Culmus resources have been [mentioned in connection with Libertinus (Serif) before][1], in one of the issues picked up in the link provide by @KrasnayaPloshchad. 

I don't know how much the design of the Hebrew and Greek (the two non-Latin sets I'm most interested in) are dear to Libertinus or Linux Libertine before that. Both are slightly compromised in a design sense (or so it seems to me), but whether "dropping in" something like Taamey Frank (from the Culmus collection) would be possible for Hebrew, I don't know (and doubt it, frankly). (The Greek is serviceable, and there would be knock-ons with Cyrillic, I expect. Best left alone, anyway.)

(P.s. Congratulations (I think!) on picking this up, @alerque! I shudder to think what this will do to your already over-burdened in-box...)

[1]: https://github.com/alerque/libertinus/issues/107#issuecomment-368026615

KrasnayaPloshchad:
Another resources available at SIL
https://software.sil.org/ezra/ezrainfo/

alerque:
Fantastic mini explanation of what the different levels of support are about: https://github.com/be5invis/Iosevka/issues/690#issuecomment-701440398

clsn:
I am willing to answer questions and blather on at length about Hebrew typography and what's needed, optional, helpful, and awesome in terms of font-support and generally attempt to be a resource (as much as I know) regarding typesetting Hebrew if I can help anyone by so doing.  I'll take a look at bugs here mentioned, maybe I can help.

alerque:
[...]

PR #492, opened 2021-12-19
work on Hebrew, vowels in Serif Regular should be ok now.

cornelisbb:
I hope I have done everything ok. This is the first time I do a pull request on github.

What I have done is improve the support for Hebrew vowels. All necessary glyphs are added to all Serif variant (except Math and Display), but they are only finetuned in Regular.

In Serif Regular vav with holam should work ok, both when the vav is used as consonant and as vowel. Also the Hebrew word for Jerusalaim should work.

I have not added support for accents and don't plan on doing so. If this is ok I will also finish the other Serif variants

Issue #340 and maybe others are addressed with this


cornelisbb:
Closing this and will do new pull request

Issue #493, opened 2021-12-21
Kerning issue with U+01DD LATIN SMALL LETTER TURNED E


reimerd:
U+01DD, the "latin small letter turned e" = `ǝ` is included in Libertinus (hurrah!). Unfortunately it has a slight kerning problem in the upright font, which is more pronounced in the italic, which (in conventional use in my world) is its typical usage (for semitic transliteration).

One would expect it to sit "neatly" in its space, but in fact it is shifted to the left (more obvious in italic than upright), so that there is a gap on the right. It is marked in this screenshot (which ironically tends to obscure the issue slightly!), but it is readily reproduced if you input some text, and copy/paste the glyph from above into it—then apply italics.

<img width="612" alt="-e-glyph" src="https://user-images.githubusercontent.com/17025551/146924600-65de4baa-0fc5-4344-b3dc-a1836deb6715.png">

I have included the same text in Brill (which is better behaved here) for comparison. Not a major issue, but would be great to have this tweaked!

Issue #486, opened 2021-08-23
Support for U+25CC ◌ DOTTED CIRCLE 


amundo:
**Describe the bug**

U+25CC ◌ DOTTED CIRCLE  is often used in IPA contexts to demonstrate dicritics in isolation.

**Expected vs. actual behavior**

Here is how the dotted circle + plus COMBINING BRIDGE BELOW renders in Gentium Plus:

<img width="615" alt="Screen Shot 2021-08-23 at 10 45 59 AM" src="https://user-images.githubusercontent.com/30924/130467638-04e08422-ccc8-4ead-8214-fff0c612deea.png">

And here is how it renders in Libertinus:

<img width="628" alt="Screen Shot 2021-08-23 at 10 45 08 AM" src="https://user-images.githubusercontent.com/30924/130467514-90fe1c0e-427e-4f5a-9b99-4f463ff2fa1d.png">

Here is that string:

◌̪

PS. Thanks for sharing Libertinus, it’s cool.

Issue #107, opened 2017-11-08
Hebrew holam not well aligned 


KrasnayaPloshchad:
In the following text, the Hebrew holam does not change its position on Vav even if you inserted ZWNJ:
```
עֲוֹנֹת עֲו‌ֹנֹת 
```
![snapshot](https://i.imgur.com/rrXQ6Y3.png)

khaledhosny:
Please give some context, I don’t know Hebrew at all.

KrasnayaPloshchad:
I don’t know Hebrew too, but as my observe, the key of this problem is the sequence ו‌ֹ (vav + ZWNJ + holam), the ZWNJ could making holam shift to left, otherwise is shift to right.

dscorbett:
The bug is that U+05B9 should be centered directly above U+05D5.

U+05B9 HEBREW POINT HOLAM is a vowel point above and to the left of its base letter. The exception is when the base letter is U+05D5 HEBREW LETTER VAV, where the holam goes directly above. However, the exception has exceptions; the holam above left on a vav is encoded as U+05BA HEBREW POINT HOLAM HASER FOR VAV. I don’t know what ZWNJ has to do with anything; maybe it was used as a workaround before U+05BA was encoded.

KrasnayaPloshchad:
The text is copied from Wikipedia article [Zero-width non-joiner](https://en.wikipedia.org/wiki/Zero-width_non-joiner), the article explains that the placement of the holam dot to the left of the letter vav is correct in Biblical Hebrew, otherwise would resemble "seasons". However it does not mentioned any more technical purposes.

mgavioli:
Can anybody with a working knowledge of Hebrew tell me if I am correct in assuming that:

1) The correct rendering of the U+05D5 + U+05B9 and of U+05D5 + U+05DA is as in the following screen shot (modified by hand from a LibreOffice screen shot by shifting slightly to the right the dot in the left-hand pair):

![issue_107_screenshot_01](https://user-images.githubusercontent.com/1420743/36414971-7d453034-1624-11e8-9a9d-2ad7183a8b59.png)

2) U+05D5 and U+05DA are the only diacritics which go **above** the HEBREW LETTER VAV.

If the answer to both points is yes, I think I can fix the issue (but I would probably lower a bit the diacritic in the U+05D5 + U+05B9 combination).

alerque:
@dajare This sounds like something you might be able to answer.

dajare:
[...]

PR #496, opened 2022-01-02
Rework of Hebrew vowels

cornelisbb:
This is a complete rework of Hebrew vowels/nikkuds/diakritics in de Serif fonts. This is done now. On my TODO list is to integrate the meteg into this, butt that is for later. All serif fonts now have proper vowel support. I have not done the sans fonts, because the Hebrew fonts there is the same as the serif.

I have also updated the fontlog with the details.

If this is accepted, I will fix the bugs on this that are reported.

The issues fixed with this PR are:
 #107 #486 #493 

I reference to the previous PR already in #340


alerque:
Sorry this has taken so long to review, I got a bit distracted and then out of the groove of font dev stuff.

I'm trying to get out a new release now and would like to include this, but there are quite a few merge conflicts with other things I've accepted in the *master* branch. Most of them I know how to handle, but the change in encodings is confusing me when trying to get the I_J stuff for dutch merged because both branches have related changes and we need most of the I_J stuff from the master branch but the encoding changes from this one to make it all jive.

I'm not sure this will make the next release window, but I'm trying.

cornelisbb:
Thanks for looking at it. I am not a git specialist, so sorry if I did something wrong. I know something about fontforge, feature files and Hebrew. I don't know how to help here, but if I can, I'll try. I can at least test the font and suggest corrections. Also for me, it has been long ago. :)

alerque:
I'll try to handle the Git stuff, I *am* an expert at that side, more so than even fonts in general and certainly more than Hebrew.

Do you recall if there was a specific reason you switched some of the font faces from UnicodeFull to UnicodeBmp while working on the Hebrew vowel stuff?

cornelisbb:
> Do you recall if there was a specific reason you switched some of the font faces from UnicodeFull to UnicodeBmp while working on the Hebrew vowel stuff?

No. As far as I've noticed this does not make a difference for Hebrew. Was probably by accident.

alerque:
Yes as far as I know it mostly affects the way the FontForge interface treats the available character set. I'm going to try to change it back for the fonts it got set for, then re-normalize and try merging the stuff from main and see if that makes the overlap in changes more manageable.

alerque:
Okay I think I got this...at least all of Git, FontForge, and FontShip are happy about the state of the source files. That does *not* mean the resulting state is correct because I had to do a lot of hand merging and I did it purely in the source using a text editor. In particular there were overlaps between this work and some other more minor fixes already in *master* and I hand merged them as best I could make sense of them.

Since I had to do this in a separate branch this is going to close as if GH things it didn't get merged, but it *should* all be in the current *master* in a moment. I'll post links to build artifacts here and if you can test that the Hebrew still works as expected I'll try to review the other changes and make sure they still work too.

alerque:
So there has been some build trouble in CI land, but I think I have it all worked out now. When [this CI run completes](https://github.com/alerque/libertinus/actions/runs/10941247810) there should be artifacts attached to it with the current font builds. I'm stepping out for a bit, but the next thing to do (assuming they actually get posted) is to test those and make sure everything is working as expected given the changes submitted in this PR.

alerque:
Somehow in merging this we lost one (Serif Regular) alternative plus signs from #504.

alerque:
I got the plus sign from #502 back just by copying the `SplineSet` back where it belongs. It appears to have been the victim of some bad merging. I hope there are not many more instances like that.

Also I fixed CI so it doesn't choke on generating the preview any more. That was always a bad rule in this project, but *fixing* a bug in Fontship made it actually start trying to work in this project, and of course fail because we don't have XeLaTeX in CI.

Build artifacts with fonts for testing are finally working, latest ones as of this comment are [on this workflow](https://github.com/alerque/libertinus/actions/runs/10949198971).

cornelisbb:
Thanks. Next week I have time to do some proper testing.

alerque:
Just so you know, I kind of have ants in my pants to tag this. That may sound strange given I had the entire project on the back burner for years and it took me ages to get around to PR reviews and conflict management, but it looks like Libertinus Serif is going to get baked into the next major release of Typst (replacing the current Linux Libertine that it embeds). I'd really like to get a new tagged release out **and** have enough days left over that it gets some wider testing before Typst 0.12 drops just to make sure there aren't serious regressions.

I don't mean to push you at all (I know all about how volunteer time management goes!) but would you perhaps have a sample text in Hebrew that particularly exercises the bits you worked on here? I'd be happy to do some of the leg work of typesetting it in a before and after version and dropping a visual diff of some kind here.

As long as we're confident the current state is an improvement, I'd even rather release a v7.050 now and have to send out a v7.051 in a week with some tidbit we found rather than not having any of the last 2 years improvements land at all.

PR #520, opened 2022-10-20
Cyrillic

StefanPeev:
It is a major update of Libertinus Serif.


imposeren:
@StefanPeev I've built your fonts and there is still an issue with letter "є" in BoldItalic version of the font: https://github.com/alerque/libertinus/issues/521

StefanPeev:
@imposeren See [v.7.052](https://github.com/StefanPeev/libertinus/tree/cyrillic/sources/UFO). In this version #521 is solved.

imposeren:
@StefanPeev the results are great, but I doubt that this pull request will be accepted as it is: too many commits without good commit messages, and looks like you also included some extra files in this MR, which I'm not 100% sure all required for building. 5k files changed looks scary. Maybe you included something that is "built by your font-editing software, but not really a source and can be rebuilt by the same software if needed".

So maybe you should:
* add some extensions to `.gitignore` file
* rebase/squash commits into single commit with more informative commit message

StefanPeev:
@imposeren Thank you! I know that the result is a revolution :) . Every single glyph in the font files was changed in one or another way. For the first time we have an option to make an interpolation between Regular and Bold, Italic and Bold Italic because their glyphs are compatible. Actually these are not changes but new beginning of the project. It is why I don't know what to do. I do not spread the OTF and TTF files of my version but it will be good if more people test the version and if they give their ideas for changes. Only after a critical observation of issued by me Libertinus Serif's files could be taken a good decision. As for me I'm absolutely sure that the direction of the project is only toward compatible files which give as an opportunity for interpolation and for working with 2 masters.

cionx:
As reported in #524, there may be a problem with the glyph for 3 (U+0033, Digit Three) in Libertinus Serif SemiBold.
The left image shows how the glyph looks like in the source file (when opened in FontForge), and the right image shows how it looks like in the built OTF-files provided at [https://github.com/StefanPeev/libertinus/releases/tag/v.75][3].

![source][1] ![build][2]

So maybe this problem is only introduced during the building process.

[1]: https://user-images.githubusercontent.com/7231796/216515142-f11075d0-e95e-4982-b72b-41c208a7da02.png
[2]: https://user-images.githubusercontent.com/7231796/216515140-b131ee87-a8a5-479d-af47-9fd29b8709f5.png
[3]: https://github.com/StefanPeev/libertinus/releases/tag/v.75

waldyrious:
For future reference regarding the closure of this PR, quoting https://github.com/alerque/libertinus/issues/524#issuecomment-1415495097: 
> I closed pull request #520 and I'll close also all updates on Libertinus Serif. The reason is that I started a project [Common Serif](https://github.com/StefanPeev/Common-Serif) which is further development of Libertinus Serif but with a lot of changes. The aim of the project is to create masters (Regular and Bold, Italic and Bold Italic), which to be compatible and by which we to have a chance to generate instances like this (Regular, Medium, SemiBold, Bold; Italic, Medium Italic, SemiBold Italic, Bold Italic). All this means a serious changes in the project, it is why I think we need a new start. If you like the idea, please, give your notices on [Common Serif](https://github.com/StefanPeev/Common-Serif) repo.

PR #555, opened 2024-09-21
Enable auto-hinted TTF builds

### LINKED ISSUE #326: Provide TTF
Word has issues when exported documents with OTF typefaces in PDF format (they don't get embedded). Would it be possible to also ship TTF in addition to the OTF files?

alerque:
Closes #326, or will if merged. For the moment I still don't see any convincing reason to enable this, but at least having a PR will enable people to download the ready-made artifacts if they want to test them out.

### DIFF
--- a/FONTLOG.txt
+++ b/FONTLOG.txt
@@ -45,6 +45,9 @@ https://github.com/alerque/libertinus
 ChangeLog
 ---------
 
+unreleased (GIT) <Libertinus> Version HEAD
+- Enable builds of autohinted static TTF variants via Fontship
+
 21 September 2024 (Caleb Maclennan) <Libertinus> Version 7.050
 - Redo Hebrew vowel marks
 - Add (for Hebrew) uniE801-3, uniFB29, shin variants, letters with dagesh,
--- a/fontship.mk
+++ b/fontship.mk
@@ -1,4 +1,3 @@
-STATICTTF =
 STATICWOFF =
 
 GSUB = sources/features/gsub.fea


alerque:
I'm a bit surprised at the reactions to this. From my point of view this is a total nothing-burger that just adds confusion to the picture because people don't know what is best to use when.

Can those that are so excited about this comment on their motivation? What does this solve for you that using the OTF builds does not? What works/looks better about TTF and in what circumstances? Have you actually test-driven the hinting here on the platform of your choice? Would disabling the hinting and just having the same font in an unhinted TTF wrapper be better for anybody?

tukusejssirs:
@alerque, while I prefer OTF to TTF, I also want to use the fonts in Google products too. I have no deep knowledge how much this PR opens the path for adding the fonts to Google Fonts, but by this PR you seem to fix the issue on which depends the addition of the fonts to GF.

Correct me if I’m wrong. :wink:

khaledhosny:
OTF is poorly supported in Windows and MS apps in general (for instance, creating PDF from MS Word using an OTF font will cause the text to be embedded as raster images!). I briefly shipped Amiri as OTF (it was always shipped as TTF) and I got a few bug reports about things breaking on Windows (for example very wide glyphs getting distorted). I eventually reverted back to releasing TTFs.

Eventually I stopped building OTFs for all my fonts altogether. This is more of a principled stance, though. `CFF` table has always been an outlier to OpenType, at the technical level it shares almost nothing with the other tables, its structure is totally alien, it parsing and processing is quite different from the rest of the font. Rendering `CFF` is also a black box. Know one knows what to do with the hinting other than Adobe, it is all magic (FreeType didn’t render `CFF` glyphs well until it got a `CFF` renderer contributed by Adobe). `CFF` is basically Adobe thing and they have an edge over everyone else, so I decided not to bother with it. `glyf` table is simpler to build, simpler to parse, and less bugs in existing software overall. Hopefully, one day we will get cubics in `glyf` as well and `CFF` will be completely absolutely redundant.

PR #556, opened 2024-09-23
Fix some composite Hebrew glyphs

cornelisbb:
In Serif Italic there were errors with some composite Hebrew fonts. Bold Italic is fine. This request fixes the issue.

### DIFF
--- a/sources/LibertinusSerif-Italic.sfd
+++ b/sources/LibertinusSerif-Italic.sfd
@@ -20,6 +20,7 @@ OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1155510122
+ModificationTime: 1727117497
 PfmFamily: 17
 TTFWeight: 400
 TTFWidth: 5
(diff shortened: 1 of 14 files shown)


sources/LibertinusSerif-Italic.sfd
```diff
@@ -20,6 +20,7 @@ OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1155510122
+ModificationTime: 1727117497
```
alerque, line 23: This needs to be normalized again to get rid of this and a few other artifacts, but I can take care of that if you don't get too it before I merge this (which should be soon, and I'll probably fast-track a patch release so folks playing around with the new Hebrew stuff have it to work on).

alerque (dismissed):

cornelisbb:
Totally forgot about normalize. I hope I have done it correctly.

alerque:
> Totally forgot about normalize. I hope I have done it correctly.

You did the right thing, but with some older bit of software. Either `fontforge` or `sfdnormalize` is older than I'm using, so it changed the SFD version back in time. **But no worries I gotcha covered.** I ran it locally and amended your commit.

I can't push it here because you opened this PR *from* your master branch, but I did merge it. In the future if you create a new branch to work on any changes GitHub will make it a little easier to collaborate because I would be able to push directly to the PR and in the end your PR would show as "merged" instead of "closed", but again not a huge deal.

Thanks for the contribution.

PR #559, opened 2024-10-08
Vertically center `integral.size1`

### LINKED ISSUE #542: low baseline in Libertinus Math  /integral.size1
Howdy. This may or may not be a bug; I'm seeking clarification.

The baseline for `/integral.size1` is low:
image

In practice in many applications, this doesn't matter.  TeX and friends encode an integral as a class 1 symbol (a large operator) and typeset it centered on the math axis when doing math layout.  MS Word appears to do the same thing.  So the output from these engines looks reasonable:
image

I was contributing recently to the math layout engine in  [Typst](https://github.com/typst/typst), which until now has relied on the baseline for large operators.  By and large, math fonts seem to set those baselines so that the large operators are centered on the axis.  The Libertinus Math `/intergral.size1` glyph seems to be exceptional in this regard, and this led to suboptimal typesetting when the baseline was relied on.
image

I'm filing this bug report to enquire if the low baseline is a bug.  It may be that there is some other context where this baseline is desired.  Or maybe the baseline is simply deemed unimportant since TeX et. al. and MS Word clobber it anyway and its current value is maybe accidental. Or maybe the baseline is an expression of what is actually desired for typesetting, though that seems unlikely.  Is there a reason the baseline for this glyph is not set such that the glyph is centered on the math axis?  Thanks!

tmke8:
Closes #542 

I used [this script](https://www.reddit.com/r/FontForge/comments/11r6x4q/vertically_centering_glyphs/llkljgq/) to center it. The script seems to have moved it down by 308. It looks correct to me, but I may be wrong.

I also don't know whether I need to change hinting somewhere? I didn't see anything obvious in the .sfd file.

This also seems to center all the other large integral glyphs (including the ss08 glyphs) because they all refer to `integral.size1`.


khaledhosny:
If we are going to do this, you have to check all big operators not just integral. It should also be centered around math axis.

tmke8:
I modified the script to center around the math axis:

```python
startglyph = 1114441
endglyph = 1114441  

thisfont =  fontforge.activeFont()  
for i in range(startglyph,endglyph+1):  
    ytop = thisfont[i].boundingBox()[-1] # counted from baseline, excluding descent!  
    ybot = thisfont[i].boundingBox()[1] # same note as above
    # Calculation method:
    # Current midpoint of glyph (counted from baseline) = (ytop + ybot) / 2
    translation = thisfont.math.AxisHeight - ((ytop + ybot) / 2)
    thisfont[i].transform( psMat.translate( 0, translation ))
```
The difference to before is 6 coordinate points.

I also investigated the other big operators (those with `.size1`  suffix in their name), but they all look centered to me:

image

Produced with this latex snippet, rendered on Temml.org

```latex
\begin{aligned}
&\cdot\int\iint\iiint\iiiint\oint\oiint\oiiint\intclockwise\varointclockwise\sum\prod\coprod \cdot\\
&\cdot\bigwedge\bigvee\bigcap\bigcup\bigcupdot\biguplus\bigsqcap\bigsqcup\cdot\bigodot\bigoplus\bigotimes\surd\Big(\Big)\Big\lceil\Big\rceil\Big\langle\Big\rangle\cdot
\end{aligned}
```

I could nevertheless run the script over these?

khaledhosny:
> I also investigated the other big operators (those with `.size1` suffix in their name), but they all look centered to me:
> 
> image

But the image shows that all integrals are not centered. You should also center the text version of the big operators (the radical and parentheses are not big operators so can be left alone).

tmke8:
> But the image shows that all integrals are not centered.

Yes... I guess I wanted to demonstrate the problem again.

Here it is with the patched font:

image


>You should also center the text version of the big operators

You mean the glyphs without `.sizeX`? Okay, I'll investigate.

tmke8:
Oh, looking at the new image, I see that the circles are too high! I'll fix it.

tmke8:
Should be fixed now:

image

And also for the slanted integrals:

image

khaledhosny:
I haven’t checked the font files, but the screenshot looks good to me.

PR #585, opened 2025-10-24
Feat/woff2 from otf

mattf1n:
Release WOFF2 files do not include smcp and c2sc font features. This PR changes the build process to instead convert the OTF files to WOFF2 so that the features are preserved.


3x10RaiseTo8:
Waiting for this PR to merge. Same woff2 files reflect in Google Fonts serving millions a font stripped of quality.

